PR #12, opened 2020-03-11
Update person.schema.json

jcolomb:
error in what is required (not everyone has a middle name)


lzehl:
Hi Julien, thanks for spotting this! This is indeed an error.
Best, Lyuba

PR #13, opened 2020-03-11
add affiliation in person metadata

jcolomb:
it would be worth looking into the PID there, Orcid is using ringold number (for me at least) but the ROR are in development. Should this be more flexible, or do you want to keep only the text version ?
Or would you like to add indexes here, and take affiliation as another json file?

### DIFF
--- a/v3.0/core/person.schema.json
+++ b/v3.0/core/person.schema.json
@@ -44,6 +44,23 @@
         ]
       }
     },
+    "affiliations": {
+      "type": "array",
+      "description": "List of institutions",
+      "items": {
+        "type": "object",
+        "properties": {
+          "name": {
+            "type": "string",
+            "description": "The full name of the affiliation (address); usually the name of an institute or company."
+          },
+          "PID": {
+            "type": "string",
+            "description": "RINGGOLD number"
+          }
+        }
+      }
+    },
     "isMainContactOf": {
       "type": "array",
       "description": "Resources for which person is the main contact.",


jcolomb:
see discussions about author list schema here https://github.com/pandoc/scholarly-metadata/blob/master/article-authoring-schema.json

lzehl:
If we keep affiliations in openMINDS they will be most likely defined in a separate schema, because one affiliation should be valid for multiple persons. Due to the current situation with GDPR it's still an open discussion though how we handle these information.

lzehl:
Thanks for your input here, though. Let's keep in touch about this.

jcolomb:
I never thought GDPR rules would apply to the researchers in this case... One way to handle this might be to ask only for orcid ID (other information can be obtained from there on demand, and orcid would deal with GDPR rights of the users) ?

skoehnen:
This could be something we should bring up in the Data Working Group Meeting. Maybe the DPO already has some thoughts about this.

skoehnen:
What should we do with this PR, are the issues already resolved?
The I would suggest we close it (The discussion will still be accessible).

lzehl:
I'll merge it in to not forget about the issue when doing the v3 schema rewrite. The GDPR issue with this addition is most likely solved in the new Agreements for using the EBRAINS/HBP platform that are currently put together. For v3 I do not see why we can not collect affiliation information for the authors.

PR #17, opened 2020-05-05
Updates to openMINDS

UlrikeS91:
Hi Lyuba,

I made some changes. Most of them are correcting typos, but some of them are suggests to be added. 

Additionally, I have some questions to some of the schemata:

dataset.schema.json
contributors (76-82): minItem and maxItem both 1? why?
ethicsApproval (85 - 88): no type? why?

file.schema.json
isCreatedFrom (52-60): description doesn’t fit

studyTarget.schema.json
isPartOf (52-58): something here doesn’t make sense. Description? anyOf?

The line numbers in brackets are the numbers in my fork (some will probably be the same in yours).
Should I turn those into issues or how would you like to handle it?

Ulrike


jcolomb:
For person.schema.json, I would keep givenname, as it is the terminology used at orcid, datacite and probably others.

UlrikeS91:
I suggested that change because I learned that given names include both first and middle names. So, if we ask for given name(s), the property middlename becomes redundant. I may have learned this wrongly, though. I'm not a native English. 
However, checking the websides you refer to, I noticed that they don't seem to solve it the way it is solved here.
ORCID: Uses first name and last name (each can have several names; and they additionally have "Published name").
Datacite: Wants family name and given name (as far as I can see this can be several for each; see their schema: http://schema.datacite.org/meta/kernel-4.3/metadata.xsd). 
So, at least these 2 don't even ask for the middle name(s) specifically. Personally, I would prefer either first, middle and last name or only given and family name.

jcolomb:
One would think people would agree on such a basic content, but they do not...

Both datacite and crossref went for given_name/givenName + surname/familyName. So getting rid of middle name might be the best way to go?

https://www.crossref.org/education/content-registration/descriptive-metadata/#00007
http://schema.datacite.org/meta/kernel-4.3/doc/DataCite-MetadataKernel_v4.3.pdf

apdavison:
It is essentially impossible to properly represent names in a simple way, see https://www.w3.org/International/questions/qa-personal-names

Probably the simplest approach is a single "preferred name" field and a series of conventions for different name forms (e.g. "first_name middle_initials last_name" for "European"-style names).

Having said that, the "family_name" and "given_name" approach is widely used (also by schema.org), with middle names/initials optionally included in "given_name"

lzehl:
Hi, sorry to drop in so late on the discussion. I agree with Andrew after giving it some thought. It seems with only given and family name we're on the safer side and more in alignment with other platforms.

jcolomb:
I would also propose to make familyName not required (like for orcid).

lzehl:
@jcolomb  For "person" we need the family and the given name, because the information is needed for being an "author" (main usage of persons in our case). I need to think about if we need to define "person" in any other context, but being an "author" or "contributor". @apdavison : what do you think?

jcolomb:
As describe in the w3 doc, family name is europe centric, what happens with people without a family name?

Dr. Julien Colomb
SFB1315
Data curation
HU Berlin
---
Private address
Schillerpromenade 4
12049 Berlin

lzehl:
Good point... never saw an author on a manuscript without one... I'll keep it in mind and consult with the curation team. I may very well follow your suggestion for not making it a required field. Thanks for this input!

lzehl (commented):
General feedback: There are too many different aspects in this pull request. For the future, it would be better to hand in a pull request for certain type of changes or changes of a specific feature. E.g., "typo corrections" could be a pull request containing (only!) typo corrections, while "dataset - source relation" could be a pull request requsting the change that a dataset "isBasedOn" / "isDerivedOn" links not only to "publicationOrSource" but directly to a dataset schema. 

feedback on changes in:  
+ v3.0/core/dataset.schema.json: too many things at once; can be integrated, but will change further  
+ v3.0/core/ethicsApproval.schema.json: can be integrated like this, but will change further  
+ v3.0/core/file.schema.json: too many things at once; can be integrated, but will change further  
+ v3.0/core/funding.schema.json: can be integrated, but will change further  
+ v3.0/core/model.schema.json: too many things at once; can be integrated, but will change further  
+ v3.0/core/person.schema.json: please change file according to the discussion for this pull request (delete middle name, stay with given and family name)  
+ v3.0/core/project.schema.json: please take this changes out of the pull request, there is (so far) no limitation on the project title or alias  
+ v3.0/core/publicationOrSource.schema.json: please correct "e.g. ..." to "e.g., ..."
+  v3.0/core/studyTarget: too many things at once; can be integrated, but will change further  
+ v3.0/core/subject.schema.json: please leave "Links to tissue sample(s) which was(were) extracted from the subject(s)." as is, (this schema will change further)
+ v3.0/core/technique.schema.json: please delete "model" and "dataset" from "produces", a technique will be restricted to link to "subject" "tissueSample" or "file" (and actually that is even wrong; update is coming soon)
+ v3.0/core/temporalSubjectInfo.schema.json: will be integrated
+ v3.0/core/tissueSample.schema.json: will be integrated
+ v3.0/options/country.schema.json: can be integrated now as reminder, will change further in future

lzehl:
@UlrikeS91 : you do not need to split up the pull request this time. I do not agree with all the changes you made, but the v3.0 schemas will change anyway a lot in the upcoming weeks and your input can be integrated as reminder on certain issues / discussions. I've pointed out some changes you should revise (see previous comment). Thanks a lot for participating in the openMINDS development! Your input is appreciated!

UlrikeS91:
> Good point... never saw an author on a manuscript without one... I'll keep it in mind and consult with the curation team. I may very well follow your suggestion for not making it a required field. Thanks for this input!

Julien is correct about the family names. Several countries don't have family names, and I know a few people that had to cope with the "European" demand of having one. However, virtually every publisher requires a family name (and I think it is similar for passports). As far as I know, they basically had to come up with something to fill in as family name (e.g. the first name twice or NFN (no first name)). 
I would probably  keep the family name as a requirement anyway. I think it is more common to have one than it is to not have one...

jcolomb:
Is there any (technical) reason to keep family name as a requirement ? If yes, is there a way to define a default one (NFN?) if nothing is entered?

UlrikeS91:
The question for me is if we need to follow specific requirements for e.g. DOIs, general publications and such. I'll try to look into this as soon as possible. If there are no specific requirements for having a first and last name for such things, then we could maybe follow Julien suggestion.

skoehnen:
What is the plan with this pull request? 
It is now almost a month old and has diverged quite a lot from upstream.

We can either close this, or if there is some more need for discussion we can convert this to a draft PR and work together on changes.
But it's not good to leave PRs open for so long, especially in a fast moving repository such as this ;)

UlrikeS91:
I think we can close this. As you say, it's out-dated and things are moving in a different direction at the moment.

skoehnen:
@UlrikeS91 ok, will do.
Just keep in mind that we always have the option to convert a pull request into a draft and work some more on it.

UlrikeS91:
@skoehnen Thanks, that's good to know such an option exists :)

PR #21, opened 2020-05-26
Max. length for character string

UlrikeS91:
Character limits for alias and headline should be define with (minLength and) maxLength (not only in description). I left out minLength because it is redunant since both properties are required.
Alias should have a maximum, too, we don't want too long aliases. Then the point of having it is lost, right?

### DIFF
--- a/v3.0/core/dataset.schema.json
+++ b/v3.0/core/dataset.schema.json
@@ -21,11 +21,13 @@
   "properties": {
     "alias": {
       "type": "string",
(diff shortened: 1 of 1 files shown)


lzehl (approved):
Good points. We might need to be less strict on the aliases (more characters) in future, but we should keep this changes as reminder for the restriction. 
We can check if there is a way of limit number of words for the abstract in the next update.

PR #22, opened 2020-05-27
dataset.schema.json update suggestions

UlrikeS91:
Additional questions: 
1. dataDescriptor: Do you intend to have the text of the data descriptor copied as a string to this field? Would this mean that figures in the data descriptor wouldn't be allowed anymore or would this be in addition to a downloadable file of it?
2. changeLog: We may want to add a field to provide a URL to the documentation for the changes (as optional additional property). For dataset, this will most oftenly not be used, but as we discussed for e.g. delineations of an atlas detailed documentation may be available.   
3. Contributors: I read up on minItem/maxItem and if I understand them correctly, you limited the contributor list to one contributor (one item). So, I removed it here. I also added uniqueItems, because I wasn't sure if this is by default true or false. Additionally, I changed the $ref from ...contributors.schema.json to ...person.schema.json. Was this on purpose and there will be a contributor schema? Was this the reason there was ´"maxItems": 1´?
4. isBasedOn: I'm lacking $ref to dataset and model here. A dataset could be based on another dataset or a model. Or is this intended to mean something else?

### DIFF
--- a/v3.0/core/dataset.schema.json
+++ b/v3.0/core/dataset.schema.json
@@ -21,11 +21,13 @@
   "properties": {
     "alias": {
       "type": "string",
-      "description": "User-defined identifier of this dataset."
+      "description": "User-defined identifier of this dataset.",
+      "maxLength": 30
     },
     "headline": {
       "type": "string",
-      "description": "Headline that describes this dataset (max. 110 characters)."
+      "description": "Headline that describes this dataset (max. 110 characters).",
+      "maxLength": 110
     },
     "abstract": {
       "type": "string",
--- a/v3.0/core/dataset.schema.json
+++ b/v3.0/core/dataset.schema.json
@@ -43,7 +43,7 @@
     },
     "changeLog": {
       "type": "string",
-      "description": "Record of all notable changes made to this dataset compared to it's previous version."
+      "description": "Record of all notable changes made to this dataset compared to its previous version."
     },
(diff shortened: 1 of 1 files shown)


skoehnen:
> 3. Contributors: I read up on minItem/maxItem and if I understand them correctly, you limited the contributor list to one contributor (one item). So, I removed it here. I also added uniqueItems, because I wasn't sure if this is by default true or false. Additionally, I changed the $ref from ...contributors.schema.json to ...person.schema.json. Was this on purpose and there will be a contributor schema? Was this the reason there was ´"maxItems": 1´?

Yes that is correct, I bet this error got stuck from main contact/custodian or something, even the description says that there should be multiple contributors.

lzehl:
@UlrikeS91 could you check if you can solve the conflict? Or let me know if this PR is obsolete and can be closed.

To your questions:
1. dataDescriptor: should be uri
2. changeLog: rename to `versionInnovations` (same as in SANDS); should be free text
3. Contributors: not sure if I understand all your points, we should discuss this again on the corresponding revision issue for this schema
4. isBasedOn: also here, we should discuss this again on the corresponding revision issue for this schema

UlrikeS91:
I think we can just ignore this. I don't remember why this was started. `dataset` doesn't exist anymore because there is `researchProduct` and `datasetVersion`. 

I close this PR.

Issue #19, opened 2020-05-10
update all $ref to resolve in jsonschema.RefResolver 


lzehl:
The schemas are defined here on the GitHub. If people are downloading the schemas in order use them for example in a Python script the $ref need to resolve in the jsonschema.RefResolver.
Solution:  
+ needs relative path?  
+ needs URL to GitHub schema file?

lzehl:
Feedback from Stefan: should resolve using the link to the GitHub file (raw).

jcolomb:
While technically easier to implement (one can get a part of the schema and get the other part directly via links to github), the link to the github file has some issues:

- dependence and persistence: what if microsoft turns github down: every link will go to a 404 (also if they decide to change the path to the raw files, actually).
- access: we saw last year that github closed its access to Iran and other countries. It was only for private repositories, but it might be different in the future. Similarly, it means that the schema will not resolve when offline.
- versioning: lets think we have version 3 and 4, links of version 3 should resolve to version 3 github files, while the master is at version 4: so links will need to be changed for each version ?

lzehl:
All good points I agree. Not sure how to proceed though. 

For versioning: in the best case I can imagine, the schemas are version dependent, but the $ref is rather independent of the version (schema should be at the same location). 

Problem at the moment: $id is currently only an identifier, but does not lead to a downloadable file.

Not sure if indirect paths could be used as $ref to get to functionally usable schemas on a local machine. 

I can't really find nice examples online, although I would have thought that this is a common problem...

lzehl:
Feedback from Tom: 
+ create a definitions.json file at the top folder (e.g., under v1.0 for MINDS) containing the $ref to all schemas (introduce short strings for shorter $ref in the schemas later)
+ use URI to GitHub raw files for $ref
+ import definitions.json into each schema, and use this to reference to other schemas 

I will follow this approach. The references in the definitions.json file are easier to maintain. If we stumble upon another solution we can modify it again.

lzehl:
@UlrikeS91 : thanks for helping. Here what needs to be done (v1.0):
- use $id of schemas in definitions.json for referencing between schemas
1) add "defintions" to each schema (see activity for example)
2) change $ref in each schema to $id of definitions (see activity for example)

UlrikeS91:
Hi @lzehl,

I added `"definitions"` to all schemata (each as it's own pull request). For some of them, I had some additional comments. Here a summary:

[...]

PR #24, opened 2020-05-29
Update dataset.schema.json

UlrikeS91:
in relation to #19 

Additional remarks: 
- Line 45: ref to `"https://schema.hbp.eu/minds/core/project.schema.json"` --> 1. problem: project.schema.json doesn't exist (it should have been placomponent according to definitions file); 2. problem: property here is `"activity"`, not project/placomponent.
I added `"#core/activity"` as I think this should be the correct reference. 
- Line 65: ref was `https://schema.hbp.eu/minds/options/license.schema.json` but schema is called `.../licensetype.schema.json` in definitions file
- Line 90: What to do for files ref? No schema/not part of minds and therefore not included in definitions. Leave it be?
- Line 97: ref was `https://schema.hbp.eu/minds/core/formats.schema.json` but schema is called `.../format.schema.json` (without the s) in definitions file
- Line 133: What to do for doi ref? No schema/not part of minds and therefore not included in definitions. Leave it be?
- Line 139 & 146: What to do for parcellationatlas & parcellationregion ref? No schemata and therefore not included in definitions. Leave it be? But they are actually part of minds (according to ref). Add schemate for them to v1?

### DIFF
--- a/v1.0/core/dataset.schema.json
+++ b/v1.0/core/dataset.schema.json
@@ -4,6 +4,9 @@
   "$schema": "http://json-schema.org/draft-07/schema#",
   "description": "Structured information describing a collection of human and/or animal data.",
   "type": "object",
+  "definitions": {
+    "$ref": "https://raw.githubusercontent.com/HumanBrainProject/openMINDS/master/v1.0/definitions.json"
+  },
   "additionalProperties": false,
   "required": [
     "@type",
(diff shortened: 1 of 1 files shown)


lzehl (approved):
Feedback:
- line 45: property "activities", previous `$ref` to "project" was a copy/past error that wasn't corrected yet, your changes are correct
- linie 65: property "license" should indeed `$ref` to "licensetype", although the property name is only `license`
- line 90: thanks for detecting this, I'll take care that this part will be updated
- line 97: your changes are correct, schema is called "format.schema.json", only the property is called "formats"
- line 133: thanks for detecting this, I'll take care that this part will be updated
- line 139 & 146: thanks for detecting this, I'll take care that this part will be updated

PR #51, opened 2020-06-02
prototype of collection-generator with methods and dynamic docstrings

lzehl:
I've updated the collection-generator to cover at least a prototype of dynamic docstrings for the dynamically compiled schema methods. 

Possible already now (code example snippets):
```python
# initiate class instance
test = CollectionGenerator('v1.0', 'test')
# print docstr for class instance 
# (incl. all dynamically built schema methods + arguments)
print(test.__doc__) 
# print docstr for schema method 
# (incl. all dynamically built instructions for a the schema method)
help(test.schemas.CORE_PERSON) # prints 
# returns MINDS 'conform' dict of a schema
lz = test.schemas.CORE_PERSON(
    at_type="https://schema.hbp/minds/Person",
    at_id="minds/core/person/v1.0.0/ze-ly.json",
    name="Ze, Ly",
    shortName="Ze, L.")
```
I want to push this to the main repo, because it might be already useful for others.

Missing (work in progress):
+ helper functions for validating / storeing collection
+ helper function for building collection?

### DIFF
--- a/collection-generator.py
+++ b/collection-generator.py
@@ -67,7 +67,6 @@ class CollectionGenerator:
                     
             # open openMINDS json-schema to find out it's properties
             with open(rp2s, 'r') as fp:
-                print(rp2s)
                 jschema = json.load(fp)
             fp.close()
             
@@ -80,6 +79,8 @@ class CollectionGenerator:
             jschema['required'].append('at_id')
             
             # create method from schema using warlock
-            setattr(self.schemas, sn, warlock.model_factory(jschema)) 
-
-        
+            setattr(self.schemas, sn, warlock.model_factory(jschema))
+            
+            # create docstring for each schema-method
+            sm = getattr(self.schemas, sn)
+            sm.__doc__ = """This is the documentation of %s""" % sn
--- a/collection-generator.py
+++ b/collection-generator.py
@@ -6,6 +6,7 @@ Created on Tue May 19 11:22:51 2020
 @author: zehl
 """
 import json
+import inspect
 import warlock
 import collections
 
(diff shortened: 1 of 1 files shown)


jcolomb:
hi, I am wondering if it would not be better to keep python scripts in a submodules/different repository, as the type of updates and commits might be different, and we may need to get scripts working on a specific release of the schema, without having to deal with versions between release (i.e. the master).

lzehl:
@jcolomb : good comment but already in the planning :) we want to keep it in here for now, but if the module becomes bigger it will get its own repo.

jcolomb:
nice.
It recently got aware of submodules functionalities (see here for a nice explanation: https://github.blog/2016-02-01-working-with-submodules/) which can be a nice solution to keep things both separate and together :)

lzehl:
nice :) This is a really neat option! I'll need to discuss it with some people, but it seems to be a good alternative to a completely separate repository. Thanks for the hint!

skoehnen:
Here is the current version of the submodules documentation:
https://git-scm.com/book/en/v2/Git-Tools-Submodules

In which direction would you create the submodule?
I think the schema should be a submodule for the python scripts.

skoehnen (approved):
Looks good to me

lzehl:
@skoehnen : which is the submodule is a good question :) need to think about it, but I think you are right. 

I'm currently trying to delete again the warlock module. It's a nice package but not completely fulfilling our needs. The core functionality is rather easy to implement (although I like it to be checked by someone with more programming experience). I'm looking at the validation and the docstring building next. 

Should we set up the submodule already for the next commit of the collection-generator? 
If yes, could we brainstorm about the module and class names (etc)?

skoehnen:
I will open an issue where we can discuss this.
We can also do a quick call if you want.

skoehnen:
Issue for python scripts in this repository #53

PR #54, opened 2020-06-03
Start example gha workflow

skoehnen:


skoehnen:
Can't test this in a different branch.
This currently just tests the python files and not the schema.
Also I am aware that it throws a lot of errors right now, sorry about that.

PR #56, opened 2020-06-03
Fix validation of tissue sample

skoehnen:
Added missing item.
Please check if I accidentally broke something.


UlrikeS91 (approved):
Your fix works as far as I can see, but some additional commetns to the schema: 

L26: "description": "Organ from which the tissue sample was extracted or which **is** represents.", --> should be "it represents"

L89: for "isDerivedFrom": Should maybe add "anyOf" for the items as well. Only one without it. Any reason not to have it?

skoehnen:
@UlrikeS91 thanks, I copied your remarks to issue #59 
Will merge this fix now and start a new PR for #59

PR #58, opened 2020-06-03
Fix validation of file

skoehnen:
I rewrote the URL property.

@lzehl please check if this is correct.


skoehnen:
After discussion with @lzehl we came to the conclusion that this it not the correct way to handle the URLs.
I will start a new pull request.

PR #63, opened 2020-06-22
Make labels compatible with existing terminologies

apdavison:
This is not intended to be merged (at least not yet), but is intended as a discussion point.

In the interests of making openMINDS as interoperable as possible with other semantic databases,  I guess we agree we should re-use existing terminologies as much as possible.

I've made a first pass for terms that seem similar to terms in [W3C PROV](https://www.w3.org/TR/prov-o/).

### DIFF
--- a/v3.0/core/dataset.schema.json
+++ b/v3.0/core/dataset.schema.json
@@ -93,7 +93,7 @@
         {"$ref": "https://schema.hbp.eu/minds3.0/core/funding.schema.json"}
       ]
     },
-    "isNewVersionOf": {
+    "wasRevisionOf": {
       "description": "Link to previously released version of this dataset.",
       "oneOf": [
         {"$ref": "https://schema.hbp.eu/minds3.0/core/dataset.schema.json"}
@@ -126,7 +126,7 @@
         ]
       }
     },
-    "isBasedOn": {
+    "wasDerivedFrom": {
       "type": "array",
       "description": "Lists the publications/sources from which this content is derived or from which it is a modification or adaption.",
       "items": {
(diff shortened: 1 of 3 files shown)


lzehl:
Hi. Good point, although all of the used once so far should have been the relations used by DataCite already (hope there is not a mismatch). Why would you not use those? Should the DataCite relations not be accepted also in W3C PROV?

Note: all v3 schemas are not up to date and will be majorly rewritten based on the last discussions on the concepts of the schemas. I'm currently collecting all of the feedback and designing the concept in the correponding wiki page (I'm too slow) ... The current implementation gives us though a lot of discussion ground. So please keep commenting, but do not put too much work in rewriting them, since they might change anyway.

apdavison:
> but do not put too much work in rewriting them, since they might change anyway

good to know! I suspected that might be the case given the mismatch between the schemas and the wiki page.

apdavison:
I don't see the terms `isBasedOn` and `isNewVersionOf` in the Datacite schema, but maybe I'm looking in the wrong place.

lzehl:
DataCite relationTypes for relatedIdentifier:
https://support.datacite.org/docs/relationtype_for_citation

This is at least what I discussed with Tom to use. Which one do you suggest?
(I don't have a strong opinion; just wanted to use well defined terms)

apdavison:
Thanks for the pointer. It's a bit of a shame that they didn't build on the PROV terms, since there's a lot of overlap, e.g. `IsDerivedFrom` and `wasDerivedFrom`. I wonder if Oli has some clever way we can use both identifiers for the same relationship without explicitly duplicating the link (e.g. in queries). This might help to integrate the Tier 1 and Tier 3 annotations.

skoehnen:
> This is not intended to be merged (at least not yet), but is intended as a discussion point.

I converted this to a draft until we either are in a state to merge or close this pull request.

lzehl:
This PR is outdated considering the schemas, but can be further discussed for the vocab (mapping to other terminologies can be defined there)

PR #68, opened 2020-08-12
product.schema.json (updated v3)

UlrikeS91:
Hi @lzehl,

I've created the first schema (product). It would be great if you could check it and see if it is correct, especially `"hasVersions"`. I wasn't 100% how to express the restriction that it can only have one type of "productVersion" but several of that one type. Additionally, it's possible I misinterpreted the concept: Should there be separate schemas for each "productVersion" (3 schemas - "datasetVersion", "modelVersion" and "softwareVersion") or one schema called "productVersion" with all 3 productVersion types in it?

Additionally, we should create the definitions files for v3 (as for v1). Would you like me to make one and you revise or would you like to give me an outline and I complete it?
You should check those in the schema already (correct use not for correct terms), and if I filled in `@type`, `@id` and `$id` correctly. 

I won't start on another schema before you checked these general things (to avoid extra work if there are mistakes).


UlrikeS91:
Updated pull request, see #69 --> closed this one.

Issue #74, opened 2020-10-12
v3 schema-revision: datasetVersion.schema.json


lzehl:
The `datasetVersion.schema.json` will contain mainly bibliographical metadata on data structures originating from human or animal experiment or a simulation.
Note: in MINDS (v1), the related schema was defined as `dataset.schema.json`.  

According to the current documentation, this schema will have the following properties:
- **`@type`** [expects: constant (`"https://openminds.ebrains.eu/core/datasetVersion"`), count: 1]
- **`@id`** [expects: free text, count: 1]
- **alias** [expects: free text, count: 1]
- **authors**: [expects: person.schema.json|organization.schema.json, count: 1 - N] 
- **availability** [expects: controlledTerm.schema.json, count: 1] 
- **custodians**: [expects: person.schema.json|organization.schema.json, count: 1 - N] 
- **otherContributions**: [expects: contribution.schema.json, count: 0 - N]
- **digitalIdentifier** [expects: digitalIdentifier.schema.json, count: 1]
- **documentation**: [expects: digitalIdentifier.schema.json, count: 1]
- **ethicsAssessment**: [expects: controlledTerm.schema.json, count: 1] (updated according to discussion below)
- **funding** [expects: funding.schema.json, count: 1 - N]
- **isNewVersionOf**: [expects: datasetVersion.schema.json, count: 0 - 1]
- **isSupplementTo**: [expects: datasetVersion.schema.json, count: 0 - N]
- **isVariantFormOf**: [expects: datasetVersion.schema.json, count: 0 - N]
- **homepage** [expects: string (format: uri), count: 0 - 1]
- **license** [expects: license.schema.json, count: 1]
- **modality** [expects: controlledTerm.schema.json, count: 1 - N] 
- **name** [expects: free text, count: 1]
- **relatedPublications**: [expects: digitalIdentifier.schema.json, count: 0 - N]
- **releaseDate** [expects: string (format: date), count: 1]
- **repository**: [expects: fileRepository.schema.json, count: 1]
- **studyTargets**: [expects: studyTarget.schema.json, count: 1]
- **summary** [expects: free text, count: 1]
- **type** [expects: controlledTerm.schema.json, count: 1 - N] 
- **versionIdentifier** [expects: free text, count: 1]
- **versionInnovations** [expects: free text, count: 1]

Note for controlledTerm.schema.json for...
[...]

PR #88, opened 2020-10-15
Update datasetversion.schema.json

UlrikeS91:
based on #74

### DIFF
--- a/v3.0/products/datasetVersion.schema.json
+++ b/v3.0/products/datasetVersion.schema.json
@@ -123,7 +123,7 @@
       "format": "date"
     },
     "repository": {
-      "description": "Link to data repository of this dataset version."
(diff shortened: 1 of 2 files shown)


skoehnen (approved):
Looks good to me.

lzehl:
@UlrikeS91 Next PR for this one: the schema filename should be "datasetVersion.schema.json" and some of the implementation needs to be adopted to the discussion/comments in #74.

UlrikeS91:
I will update it now based on the comments in #74 and I will already change `ethicsApproval` to `ethicsAssessment` (with `controlledTerm`as object and count = 1). I cannot see a reason Will/Ida would have anything against this, and even if they do, it's easy to change it again ;)

lzehl:
I agree! Go ahead!

PR #91, opened 2020-10-16
Add missing quote and comma sign

skoehnen:


skoehnen:
@UlrikeS91 I set the type of type to string for now, so the validation does not complain.

lzehl:
looks good

PR #92, opened 2020-10-16
Fix wrong format of string

skoehnen:


skoehnen:
Did just a quick fix of a deprecated file to get validation to work

Issue #85, opened 2020-10-13
v3 schema-revision: funding.schema.json


lzehl:
The `funding.schema.json` will be used to state the funding of the connected research product version.  

According to the current documentation, this schema will have the following properties:
- **`@type`** [expects: constant (`"https://openminds.ebrains.eu/core/funding"`), count: 1]
- **`@id`** [expects: free text, count: 1]
- **acknowledgement** [expects: free text, count: 0 - 1]
- **awardNumber** [expects: free text, count: 1 ?]
- **awardTitle**: [expects: free text, count: 0 - 1 ?]
- **funder** [expects: organization.schema.json, count: 1] (previously "funderName", but this seemed weird not being a free text)

PR #93, opened 2020-10-20
Create funding schema

skoehnen:
For issue #85


v3.0/products/funding.schema.json
```diff
@@ -0,0 +1,39 @@
...
+  "required": [
+    "@type",
+    "@id",
+    "funderName"
+  ],
+  "properties": {
+    "@type": {
+      "description": "Metadata node type (schema).",
+      "type": "string",
+      "const": "https://openminds.ebrains.eu/core/contribution"
```
lzehl: needs to be "..../core/funding"

v3.0/products/funding.schema.json
```diff
@@ -0,0 +1,39 @@
...
+    },
+    "awardNumber": {
+      "description": "Number of this awarded funding.",
+      "type": "string"
+    },
+    "awardTitle": {
+      "description": "Title of this awarded funding.",
+      "type": "string"
+    },
+    "funderName": {
```
lzehl: I know I did use "funderName" before, but it does seem weird to me now that a property with NAME in it is not a free text. Could you replace it with "funder" ?

Description: "Link to the funding organization"

lzehl (changes requested):

v3.0/products/funding.schema.json
skoehnen: I thought that as well, but you specified `contribution` in #85 

But that is no big problem I can change it 😉

v3.0/products/funding.schema.json
skoehnen: Sure

skoehnen: I changed

v3.0/products/funding.schema.json
skoehnen: I updated the const, is this okay that way?

lzehl:
Looks good to me. thanks

Issue #84, opened 2020-10-13
v3 schema-revision: organization.schema.json


lzehl:
The `organization.schema.json` will be used to identify organizations that were involved in producing the research products or their versions.  
Note: funding is handled separately. 

According to the current documentation, this schema will have the following properties:
- **`@type`** [expects: constant (`"https://openminds.ebrains.eu/core/contribution"`), count: 1]
- **`@id`** [expects: free text, count: 1]
- **digitalIdentifier**: [expects: digitalIdentifier.schema.json, count: 0 - 1]
- **homepage**: [expects: string (format: uri), count: 0 - 1]
- **name**: [expects: string (free text), count: 1]
- **parentOrganization**: [expects: organization.schema.json, count: 0 - N] (added because of comment below)

apdavison:
What about adding a `parent` field? For example, my institute is the Paris-Saclay Institute of Neuroscience, it has two parent organisations: the CNRS and the Université Paris-Saclay. Any of these three might be interested in seeing how many datasets they have shared through EBRAINS.

lzehl:
Agreed. I'll add it to the properties. 

@apdavison what about renaming "name" into "legalName" ? Do you think that is necessary?

apdavison:
> what about renaming "name" into "legalName" ? Do you think that is necessary?

no, I don't think so

PR #94, opened 2020-10-20
Create organization schema

skoehnen:
Related to #84


v3.0/products/organization.schema.json
```diff
@@ -0,0 +1,35 @@
...
+      "description": "Metadata node type (schema).",
+      "type": "string",
+      "const": "https://openminds.ebrains.eu/core/datasetVersion"
+    },
+    "@id": {
+      "description": "Metadata node identifier.",
+      "type": "string"
+    },
+    "digitalIdentifier": {
+      "description": "???",
```
skoehnen: I am not sure what a good description would be here.
@UlrikeS91 @lzehl maybe you have some good suggestions.

What would the digitalIdentifier of an organization be?

skoehnen (commented):
The required property missing

v3.0/products/organization.schema.json
apdavison: Maybe take a look at https://ror.org and https://www.grid.ac ?

apdavison: ROR (e.g. https://ror.org/02nv7yv05) looks like the best. More information at https://scholarlykitchen.sspnet.org/2019/12/04/are-you-ready-to-ror-an-inside-look-at-this-new-organization-identifier-registry/

skoehnen: Ah interesting.
I wasn't aware of those registrars

But that makes sense to me.

v3.0/products/organization.schema.json
```diff
@@ -0,0 +1,35 @@
...
+  "$id": "https://openminds.ebrains.eu/core/organization.schema.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "description": "Identify organizations that were involved in producing the research products or their versions.",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "@type": {
+      "description": "Metadata node type (schema).",
+      "type": "string",
+      "const": "https://openminds.ebrains.eu/core/datasetVersion"
```
lzehl: this needs to be ".../core/organization"

v3.0/products/organization.schema.json
lzehl: Globally unique and (best) persistent digital identifier for the organization.

lzehl: I agree with  @apdavison here, ROR should be preferred but we will not limit it to this

lzehl (commented):
Required properties are okay for now. The globalIdentifier should be highly recommended, though.

lzehl (changes requested):
- `@type` needs to be corrected
- `globalIdentifier` `description` : I made a different suggestion

lzehl:
@skoehnen I think Ulrike's review is not needed in addition. @UlrikeS91 hope you agree.

skoehnen:
> @skoehnen I think Ulrike's review is not needed in addition. @UlrikeS91 hope you agree.

Hi, I added both of you so you know that someone can review it.
When we started with this we agreed that it is okay if one of us reviews the pull requests and I was still working under this assumption.
If that changed in the meantime, I will change how I set up the Pull Requests?

v3.0/products/organization.schema.json
skoehnen: > Globally unique and (best) persistent digital identifier for the organization.

What do you mean with best in this case?
The best identifier or the identifier should better be persistent.

In either case I would remove `(best)`

v3.0/products/organization.schema.json
skoehnen: Oh yes, you are totally correct.

skoehnen: Is fixed.

skoehnen:
> Required properties are okay for now. The globalIdentifier should be highly recommended, though.

Yes I agree, sadly we only have required 😅 
And I think to require to would maybe a bit too strict and cause problems.

skoehnen:
> * `globalIdentifier` `description` : I made a different suggestion

You called the digitalIdentifier globalIdentifier a few times now.
Should we change the name.

(I had to check #84 to make sure I didn't make an error in the first place 😅)

v3.0/products/organization.schema.json
skoehnen: @lzehl I updated the description if you want to take a look.

lzehl:
@skoehnen so sorry :see_no_evil: the property is still called `digitalIdentifier`... I just always confuse it... In principle they identifiers should be all globally unique, so we could also call the property `globalIdentifier` but I wasn't sure if "global" is really always true for all the cases this property (and corresponding schema) is used.

skoehnen:
> @skoehnen so sorry 🙈 the property is still called `digitalIdentifier`... I just always confuse it... In principle they identifiers should be all globally unique, so we could also call the property `globalIdentifier` but I wasn't sure if "global" is really always true for all the cases this property (and corresponding schema) is used.

Okay 👍
Just wanted to make sure

UlrikeS91 (approved):

UlrikeS91:
@skoehnen and @lzehl, Is there anything I need to do here? Based on the comments:

1) Yes, I agree. Only one review should be enough. So, whenever Lyuba reviews, I don't need to review as well. I also got it like Stefan explained it, and if that changed, I also didn't get that message, @skoehnen :D

2) I reviewed it anyway to see if something was missing (which would explain why the schema isn't merge into the master yet), but it looks good to me. Stefan change the description as requested by Lyuba. @lzehl, Do you agree that we should merge this now?

lzehl:
@UlrikeS91 yes go ahead and merge it :+1:

lzehl (approved):

Issue #100, opened 2020-10-22
v3 schema-revision: copyright.schema.json


lzehl:
The `copyright.schema.json` will be used to define the copyright holder and year.

According to the current documentation, this schema will have the following properties:
- **`@type`** [expects: constant (`"https://openminds.ebrains.eu/core/copyright"`), count: 1]
- **`@id`** [expects: free text, count: 1]
- **holder** [expects: person.schema.json | organization.schema.json, count: 1]
- **year** [expects: year, count: 1]

lzehl:
@jagru20 here the documentation of what needs to be implemented (connected to discussion in #76 ). Would you like to implement it or should I assign someone else?

jagru20:
I think I can do this.

lzehl:
thanks :slightly_smiling_face:

PR #105, opened 2020-10-23
Create copyright.schema.json

jagru20:
Related to #100 

As I did not find any definition of "year" in the schema, I used the RegEx pattern for this.

### DIFF
--- a/v3.0/products/copyright.schema.json
+++ b/v3.0/products/copyright.schema.json
@@ -23,7 +23,7 @@
     },
     "holder": {
       "type": "array",
-      "description": "The person or organisation that holds the copyrights.",
+      "description": "The person or organisation that holds the copyright.",
       "items":{
         "oneOf":[
           {"$ref":"https://raw.githubusercontent.com/HumanBrainProject/openMINDS/master/v3.0/definitions.json#definitions/PERSON"},
@@ -34,7 +34,7 @@
     "year": {
       "type": "string",
       "pattern":"([0-9]{4}])",
-      "description": "Indicates the year in which the copyrights apply."
+      "description": "The year during which the copyright was first asserted."
     }
   }
 }
--- a/v3.0/products/copyright.schema.json
+++ b/v3.0/products/copyright.schema.json
@@ -1,0 +1,40 @@
+{
+  "$id": "https://openminds.ebrains.eu/core/copyright.schema.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",  
+  "description": "Copyright holder and year of the research product",
+  "title": "Copyright",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "@type",
+    "@id",
+    "holder",
+    "year"
+  ],
+  "properties": {
+    "@type": {
+      "type":"string",
+      "const": "https://openminds.ebrains.eu/core/copyright",
+      "description": "Metadata node type (schema)."
+    },
+    "@id": {
+      "type": "string",
+      "description": "Metadata node identifier."
+    },
+    "holder": {
+      "type": "array",
+      "description": "The person or organisation that holds the copyrights.",
+      "items":{
+        "oneOf":[
+          {"$ref":"https://raw.githubusercontent.com/HumanBrainProject/openMINDS/master/v3.0/definitions.json#definitions/PERSON"},
+          {"$ref":"https://raw.githubusercontent.com/HumanBrainProject/openMINDS/master/v3.0/definitions.json#definitions/ORGANISATION"}
+        ]
+      }
+    },
+    "year": {
+      "type": "string",
+      "pattern":"([0-9][0-9][0-9][0-9])",
+      "description": "Indicates the year in which the copyrights apply."
+    }
+  }
+}
--- a/v3.0/products/copyright.schema.json
+++ b/v3.0/products/copyright.schema.json
@@ -33,7 +33,7 @@
     },
     "year": {
       "type": "string",
-      "pattern":"([0-9][0-9][0-9][0-9])",
+      "pattern":"([0-9]{4}])",
       "description": "Indicates the year in which the copyrights apply."
     }
   }


v3.0/products/copyright.schema.json
```diff
@@ -0,0 +1,40 @@
...
+      "items":{
+        "oneOf":[
+          {"$ref":"https://raw.githubusercontent.com/HumanBrainProject/openMINDS/master/v3.0/definitions.json#definitions/PERSON"},
+          {"$ref":"https://raw.githubusercontent.com/HumanBrainProject/openMINDS/master/v3.0/definitions.json#definitions/ORGANISATION"}
+        ]
+      }
+    },
+    "year": {
+      "type": "string",
+      "pattern":"([0-9][0-9][0-9][0-9])",
```
skoehnen: I would suggest this pattern 
`([0-9]{4})`
to simplify it a bit.

v3.0/products/copyright.schema.json
```diff
@@ -0,0 +1,40 @@
+{
+  "$id": "https://openminds.ebrains.eu/core/copyright.schema.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",  
+  "description": "Copyright holder and year of the research product",
+  "title": "Copyright",
```
skoehnen, line 5: We haven't used the title in the other schemas.
Maybe it's a good idea to add it.

In any case we should be consistent, what do you think @lzehl ?

skoehnen (changes requested):

v3.0/products/copyright.schema.json
jagru20: I have just used it, because it is also mentioned in [this](https://github.com/HumanBrainProject/openMINDS/wiki/Implementation) example, but if it is not necessary, we can omit it.

skoehnen: Okay, since it is in the implementation example we should definetly add it to the other schemas.

skoehnen: I will add an issue, so we don't forget it.

v3.0/products/copyright.schema.json
jagru20: of course :D

v3.0/products/copyright.schema.json
```diff
@@ -0,0 +1,40 @@
...
+      "const": "https://openminds.ebrains.eu/core/copyright",
+      "description": "Metadata node type (schema)."
+    },
+    "@id": {
+      "type": "string",
+      "description": "Metadata node identifier."
+    },
+    "holder": {
+      "type": "array",
+      "description": "The person or organisation that holds the copyrights.",
```
UlrikeS91: I would suggest: "The person or organisation that holds the copyright." (no plural)

v3.0/products/copyright.schema.json
```diff
@@ -0,0 +1,40 @@
...
+        "oneOf":[
+          {"$ref":"https://raw.githubusercontent.com/HumanBrainProject/openMINDS/master/v3.0/definitions.json#definitions/PERSON"},
+          {"$ref":"https://raw.githubusercontent.com/HumanBrainProject/openMINDS/master/v3.0/definitions.json#definitions/ORGANISATION"}
+        ]
+      }
+    },
+    "year": {
+      "type": "string",
+      "pattern":"([0-9]{4}])",
+      "description": "Indicates the year in which the copyrights apply."
```
UlrikeS91: I would suggest "The year in which the copyright was issued." @skoehnen, is that correct? Are copyrights issued?

UlrikeS91 (changes requested):
Just some suggestions for the descriptions, otherwise the schema looks good :)

v3.0/products/copyright.schema.json
skoehnen: I think granted is better in this context, because issued implies that some other entity vetted the copyright of the data.

You can of course register your intellectual property in some countries, but that is more of a legal tool to help you during litigation. That way you can establish a point in time where you established your copyright.
I don't think that this is necessary for our use-cases.

v3.0/products/copyright.schema.json
lzehl: I agree

v3.0/products/copyright.schema.json
lzehl: @apdavison what do you think?

jagru20: I am no expert in this, but does not "granted" also imply the verification by someone else? To me "The year for which copyright is claimed" sounds appropriate.

lzehl: From the definition point of view I think the copyright year usually corresponds to the year the work was published. 
Definition from schema.org: "The year during which the claimed copyright for the CreativeWork was first asserted." I suggest to stay close to this one.

UlrikeS91: I found something similar: "The first year of publication of the copyright.". 

What about: **"The year during which the copyright was first asserted."**?

jagru20: Sounds good.

lzehl: :+1:

v3.0/products/copyright.schema.json
```diff
@@ -0,0 +1,40 @@
...
+      "items":{
+        "oneOf":[
+          {"$ref":"https://raw.githubusercontent.com/HumanBrainProject/openMINDS/master/v3.0/definitions.json#definitions/PERSON"},
+          {"$ref":"https://raw.githubusercontent.com/HumanBrainProject/openMINDS/master/v3.0/definitions.json#definitions/ORGANISATION"}
+        ]
+      }
+    },
+    "year": {
+      "type": "string",
+      "pattern":"([0-9]{4}])",
```
jagru20, line 36: One thing just came to my mind: How would we deal with cases where copyright is intended to be a span of years? We could make the year property an array containing each individual year within the range, or we could use the `oneOf` operator to have the choice between a pattern for a single year and a range of years.

jagru20: Or do we simply always choose the first year (as we just decided to write in the description)?

lzehl: first year as in the discription. The copyright span is counted from that year on.

lzehl (approved):

UlrikeS91 (approved):

lzehl:
@UlrikeS91 or @skoehnen feel free to merge this one

PR #117, opened 2020-10-26
Update softwareversion.schema.json

jagru20:
Added description


v3.0/products/softwareVersion.schema.json
```diff
@@ -1,7 +1,7 @@
 {
   "$id": "https://schema.ebrains.eu/openminds/softwareversion.schema.json",
   "$schema": "http://json-schema.org/draft-07/schema",
-  "description": "",
+  "description": "Structured meta data for this software version",
```
lzehl: I prefer and use "metadata" instead of "meta data"

lzehl (changes requested):

lzehl (approved):

Issue #119, opened 2020-10-26
v3 schema-revision: softwareFeature.schema.json


lzehl:
The `softwareFeature.schema.json` will contain metadata on the software feature.

According to the current documentation, this schema will have the following properties:
- **`@type`** [expects: constant (`"https://openminds.ebrains.eu/core/softwareFeature"`), count: 1]
- **`@id`** [expects: free text, count: 1]
- **definition**: [expects: string (format: iri, count: 0 - 1] 
- **description**: [expects: free text, count: 1] 
- **category**: [expects: controlledTerm.schema.json, count: 0 - 1] 
- **name** [expects: free text, count: 1]

Note for controlledTerm.schema.json for...
- **category** (`softwareFeatureCategory`): expects JSON-LDs for (cf. uniMINDS entries?)

lzehl:
@jagru20 could you take this over and also see if something is missing / misunderstood in the documentation?

jagru20:
yes, do you assign me?

PR #128, opened 2020-10-26
Create softwareFeature.schema.json

jagru20:
#119

### DIFF
--- a/v3.0/products/softwareFeature.schema.json
+++ b/v3.0/products/softwareFeature.schema.json
@@ -33,7 +33,7 @@
     "definition":{
       "type": "string",
       "format" : "iri",
-      "desription" : "Identifier (IRI) to definition of the feature." 
+      "description" : "Identifier (IRI) to definition of the feature." 
     },
     "category":{
       "description":"Category that this feature belongs to.",      
--- a/v3.0/products/softwareFeature.schema.json
+++ b/v3.0/products/softwareFeature.schema.json
@@ -24,7 +24,7 @@
     },
     "name": {
       "type": "string",
-      "description": "The name of this software feature.",
+      "description": "The name of this software feature."
     },
     "description": {
       "type": "string",
--- a/v3.0/products/softwareFeature.schema.json
+++ b/v3.0/products/softwareFeature.schema.json
@@ -2,7 +2,7 @@
   "$id": "https://openminds.ebrains.eu/core/softwareFeature.schema.json",
   "$schema": "http://json-schema.org/draft-07/schema#",  
   "description": "Feature included in this software",
-  "title": "Software Feature",
+  "title": "software feature",
   "type": "object",
   "additionalProperties": false,
   "required": [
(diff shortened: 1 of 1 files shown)


lzehl:
@jagru20 I changed it to a draft or is it ready for the review?

lzehl:
@jagru20 are you done with this? should I start the review?

jagru20:
Yes, I think it is ready to be reviewed...

v3.0/products/softwareFeature.schema.json
```diff
@@ -0,0 +1,43 @@
+{
+  "$id": "https://openminds.ebrains.eu/core/softwareFeature.schema.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",  
+  "description": "Feature included in this software",
+  "title": "Software Feature",
```
lzehl: I would always use the label used for schema also for the title: meaning "softwareFeature"
Or we choose a display label which I would not capitalize: "software feature"

v3.0/products/softwareFeature.schema.json
```diff
@@ -0,0 +1,43 @@
...
+    },
+    "name": {
+      "type": "string",
+      "description": "The name of this software feature."
+    },
+    "description": {
+      "type": "string",
+      "description": "The description of this software feature."
+    },
+    "definition":{
```
lzehl, line 33: What is the definition? Is this a link to an ontology term? Could you explain a bit more?

lzehl (changes requested):
"request change" is misleading but I would like to clarify my questions/suggestions first

v3.0/products/softwareFeature.schema.json
jagru20: Ok I was also wondering if this was correct. I thought it was rather meant to be a display label, because it is what shows up in the KG editor later on? However, of course we can make it not capitalized.

v3.0/products/softwareFeature.schema.json
jagru20: `definition` is a link to an external resource, defining the specific feature. For example the value of `definition` for the feature "command line interface" would be https://www.wikidata.org/wiki/Q189053

I could change the description of the `definition` property to "**Link to external resource defining the feature**" to make this clear.

lzehl: Mh... this one should then maybe be renamed. Let's do that though in another PR.

v3.0/products/softwareFeature.schema.json
lzehl: I chatted with @olinux in the meantime. Go for a display label, but I suggest do not capitalize it "software feature"

v3.0/products/softwareFeature.schema.json
jagru20: You mean the renaming, but I change the `description` of definition in this one, right?

Issue #131, opened 2020-10-28
refinements of schema format


lzehl:
Based on discussions with @olinux we're going to simplify the used schema format and provide a generator.py (somewhere) that is going to translate those user-friendly reduced schema formats to proper json-schema (or far future: also other schema formats).

Affected by this is mainly the general implementation of a schema on the Githubs (openMINDS and SANDS), in particular how to reference between schemas. Indirectly this will require an update also on the check performed on the schemas on the Githubs.

NEW schema implementation template:
```json
{
  "_type": "https://openminds.ebrains.eu/SUBDOMAIN/SCHEMALABEL",
  "description": "One sentence description of what metadata this schema covers.",
  "required": [
    "PROPERTYLABEL"
  ],
  "properties": {
    "PROPERTYLABEL": {
      "type": "TYPE",
      "_instruction": "Instruction on what metadata value is expected here."
    },
    "PROPERTYLABEL": {
      "type": "string" OR "array",
      "_instruction": "Enter link to other openMINDS metadata file of the stated schema type.",
      "_linkedTypes": [
        "https://openminds.ebrains.eu/SUBDOMAIN/SCHEMALABEL"
      ]
    }
  }
}
```

Further instructions / explanations:
- **`"_type"`** will replace (combine) the previous **`"$id"`** and **`"@type"`**.
- **SUBDOMAIN options** (with incomplete tier 3 set):
  - `"core"` (tier 1)
  - `"sands"` (tier 2)
  - `"electrophysiology"` (tier 3)
  - ??? (tier 3)
- **SCHEMALABEL** will be capitalized (example: **`"ResearchProduct"`**).
- **`"_linkedTypes"`** will replace the previous **`"$ref"`**
  - if a single link is expected for a **`"_linkedTypes"`** property use `"type": "string"`
  - if multiple links are expected for a **`"_linkedTypes"`** property use `"type": "array"`
  - in case of **`"_linkedTypes"`**, additional keys for properties of `"type": "array"` are the same as in json-schema without defining **`"items"`**
  - note: within an array items have to be of the same type (schema types listed in **`"_linkedTypes"`** means OR only)

Note: A definitions.json is not gonna be needed any more after this change. Also not needed anymore for the schema implementation is the **`"@id"`** and **`"@type"`** (they will be taken care of "globally" by the generator).

[...]

Issue #127, opened 2020-10-26
v3 schema organization


lzehl:
I would like to meaningful organize the schemas within openMINDS v3.0.
Suggested structure (last update Oct 27):

**products/**
    ~`contribution.schema.json`~ (moved to miscellaneous)
    ~`copyright.schema.json`~ (moved to miscellaneous)
    `datasetVersion.schema.json`
    ~`digitalIdentifier.schema.json`~ (copy/paste error; was moved to miscellaneous)
    ~`funding.schema.json`~ (moved to miscellaneous)
    ~`license.schema.json`~ (moved to miscellaneous)
    `metadataConstructVersion.schema.json`
    `modelVersion.schema.json`
    `project.schema.json`
    `researchProduct.schema.json`
    `softwareFeature.schema.json`
    `softwareVersion.schema.json`

**data/**
    `fileBundle.schema.json`
    `fileInstance.schema.json`
    `fileRepository.schema.json`
    `hash.schema.json`
    `mediatype.schema.json`

**experiment/**
    `parameterSetting.schema.json`
    `protocol.schema.json`
    `protocolExecution.schema.json`
    `studyTarget.schema.json`

**specimen/**
    `subject.schema.json`
    `subjectState.schema.json`
    `tissueSample.schema.json`
    `tissueSampleState.schema.json`

**miscellaneous/**
    `contribution.schema.json`
    ~`controlledTerm.schema.json`~ (moved to controlledTerms; based on discussion with Oli)
    `copyright.schema.json`
    `digitalIdentifier.schema.json`
    `funding.schema.json`
    `license.schema.json`
    `organization.schema.json`
    `person.schema.json`
    `resource.schema.json`

**controlledTerms/**
    `controlledTerm.schema.json`
    `<<TERM-TYPE>>.schema.json` (e.g., biologicalSex.schema.json)

lzehl:
@UlrikeS91 @apdavison @bweyers @olinux @skoehnen it would be great to hear your opinion on this.

@olinux let me know if the substructure somehow should be more useful for the KG team (e.g. for handling the different KG spaces) and needs adaptions for this purpose

UlrikeS91:
Overall, agreed. The folders seem reasonable to me. Just a few things I'm unsure about:

1. `digitalIdentifier` is listed twice. Pretty sure you only want to have it under miscellaneous/ and not under products/. If you intended to have it under products/, I would disagree ;) 
[...]

PR #132, opened 2020-11-02
restructuring of schema repository according to issue #127

lzehl:
This PR restructures the schemas in the repository of v3.0 according to issue #127 

NOTE: these changes will break the references in the definitions.json, but since this file will be most likely deleted in an upcoming PR this should be ignored here (cf. #131)


skoehnen (approved):
Looks good to me

PR #138, opened 2020-11-04
Update of all schemas for openMINDS core v3.0

olinux:


lzehl:
@olinux thanks a lot! I'll merge it directly and then go through it step by step tomorrow to correct the small issues we discussed in the last days.

PR #144, opened 2020-11-27
Add contribution guideline

skoehnen:
I moved your (@lzehl) text from README.md to the CONTRIBUTING.md file, update the text a bit and added some more information.


lzehl (changes requested):
The how to contribute is not that long so I would leave it in the README.

lzehl:
@skoehnen never mind the last request.

PR #145, opened 2020-11-27
Add contribution guide document

skoehnen:
I moved your (@lzehl) text from README.md to the CONTRIBUTING.md file, update the text a bit and added some more information.


lzehl (changes requested):
I would leave it in the readme. its not that long.

skoehnen:
The advantage of a dedicated CONTRIBUTING document is that it is integrated in the github process to create a pull request.

lzehl:
@skoehnen never mind then the request. thanks!

PR #160, opened 2021-02-05
Update subject.schema.tpl.json & tissueSample.schema.tpl.json

lzehl:
Adding "isPartOf" to link subject into a subject group and tissue sample into tissue sample collection.


UlrikeS91 (approved):
I like this. Make the structure a bit more flexible and allows stating all subjects while still grouping them into groups together :)

Issue #158, opened 2021-02-05
Add affiliations to link Person and Organization


apdavison:
In the context of Live Papers, I would like to be able to provide affiliations for the authors. The LivePaper schema should probably be in an openMINDS extension, but I think the concept of affiliation is sufficiently generic to belong in core.

Specifically, I propose to add an Affiliation schema in core/actors, which would link one Person with one Organization, and would have a start date and end date as optional additional fields.

lzehl:
I agree that affiliations should be covered in the core (we had it on the list and then forgot about it again...). 
Do we need to be able to hide them like the emails? 

TODO then: add a property called "affiliation" to Person schema that links to an Affiliation schema. 
Adding an Affiliation schema to ACTORS with properties "startDate" (type: string, format: date), "endDate" (type: string, format: date) and "organization" (type: object, link to Organization schema).

@apdavison Do you agree? Could you take care of the PR or should I?

apdavison:
@lzehl I'll make a PR, it will be good practice for getting to grips with the schema templates

lzehl:
great :slightly_smiling_face: let me know if you have questions. I still need to do a proper documentation for the template dialect.

lzehl:
This issue was solved in the PR #162. I'll close it now.

PR #162, opened 2021-02-05
Add Affiliation schema

apdavison:
cf #158

### DIFF
--- a/schemas/actors/person.schema.tpl.json
+++ b/schemas/actors/person.schema.tpl.json
@@ -30,6 +30,7 @@
     },
     "affiliation": {
       "type": "array",
+      "minItems": 1,
       "uniqueItems": true,
       "_instruction": "Add the current and, if necessary, past affiliations of this person",
       "_linkedTypes": [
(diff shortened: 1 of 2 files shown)


schemas/actors/person.schema.tpl.json
```diff
@@ -27,6 +27,14 @@
     "givenName": {
       "type": "string",
       "_instruction": "Enter the given name of this person."
+    },
+    "affiliation": {
+      "type": "array",
+      "uniqueItems": true,
```
lzehl, line 34: This needs a `"minItems": 1` as well, because otherwise an empty array can be provided. We decided to force to avoid empty arrays for optional properties (because they can be left out completely if no values can be provided).

lzehl (changes requested):
Thanks a lot for taking care of this PR. There is only a small addition needed.

schemas/actors/person.schema.tpl.json
apdavison: done

PR #168, opened 2021-02-16
Introduction of "_formats" and "_categories"

### LINKED ISSUE #159: Add contactInformation-alike for softwareVersion or researchProductVersion
In a discussion on how to map the former `help` -Attribute onto the openMINDS schema, @olinux and I have been struggling to find a proper equivalent. `help` can contain several entries, from which each can be a url or a mail address or similar. This is meant as an answer to a possible user's question "How can I do XY with this software?" assuming, that the answer is not in the documentanion. 
We considered `otherContributions` in combination with the contributionType `softwareSupport` but the problem here is that it cannot store information other than that there was a contribution and by whom, not where to find it. 
To solve this, we suggest adding an optional property to `softwareVersion `or `researchProductVersion` that stores this information in a string.

lzehl:
Introduction of "_formats" to simplify the dialect for stating multiple formats for properties of type "string". 
Connected to #159


lzehl:
@olinux I think I've changed now all "format" property keys to "_formats": [] for all properties of type "string" within openMINDS core. I will next change it in openMINDS SANDS and controlledTerms. Are you taking over the necessary changes in the openMINDS generator?

lzehl:
@olinux I've also introduced now first examples of "_categories" and "_linkedCategories" (case one: "studyTarget", case two: "legalPerson"). I hope this is enough to test the corresponding changes in the openMINDS generator for these new schema.tpl.json key words. Let me know if you need more. The other repos (SANDS, controlledTerms) are also ready to go with the related PRs.

olinux (approved):

PR #177, opened 2021-03-03
Changing / Extending ParameterSetting

### LINKED ISSUE #161: Changing / Extending ParameterSetting
This issue is made for documentation purposes and to incorporate more feedback (if necessary).
It was discussed with @UlrikeS91 and @olinux offline in advance.

**Summary of the offline discussion:**

First, the current ParameterSetting schema is meant to only hold one parameter setting. Discussing more use cases / examples this turns out to be not sufficient in case of parameter settings that are strongly related (e.g., `"anestheticAgent": "Isoflurane"` and `"concentration": "0.5%"` (the latter of course as QuantitativeValue)). 

This could be handled either by introducing a "isRelatedTo" in the ParameterSetting schema (which may lead quickly to rather complex structures) or by grouping parameter settings. 

Second, the properties of the parameter settings are misleading at the moment. A change of the properties should be considered.

And third, would it not be smart to enter a QuantitativeValue for numeric parameters? Allowing an entry to be an object OR a string seems to be technically difficult. Therefore this part cannot be implemented. 

**Summary of suggested solution:**

Two schemas: ParameterSettings schema which is grouping (embedded) Parameter schema.

**`ParameterSettings.schema.tpl.json` properties:**
+ "parameters" (type: object (as embeddedTypes), link to Parameter.schema.tpl.json, count: 1-n)
+ "relevantFor" (type: object, link to BehavioralTask OR Technique, count: 1)

**`Parameter.schema.tpl.json` properties:**
+ "parameterName" (type: string, free text)
+ "parameterEntry" (type: string OR number)
+ "unitOfMeasurement" (type: object, link to UnitOfMeasurement)
+ "description" (type: string, free text)

@apdavison , @UlrikeS91 , @olinux would that work? what do you think?

@apdavison would that collide with the in-depth schemas?

lzehl:
The changes in this PR were discussed in issue #161

### DIFF
--- a/schemas/research/numericalParameter.schema.tpl.json
+++ b/schemas/research/numericalParameter.schema.tpl.json
@@ -13,7 +13,7 @@
       "type": "array",
       "minItems": 1,
       "uniqueItems": true,
-      "_instruction": "Add at least one quantitative value for this numerical parameter.",
+      "_instruction": "Add at least one quantitative value for this parameter.",
       "_linkedTypes": [
         "https://openminds.ebrains.eu/core/QuantitativeValue",
         "https://openminds.ebrains.eu/core/QuantitativeValueRange"
--- a/schemas/research/numericalParameter.schema.tpl.json
+++ b/schemas/research/numericalParameter.schema.tpl.json
@@ -14,7 +14,7 @@
       "minItems": 1,
       "uniqueItems": true,
       "_instruction": "Add at least one quantitative value for this parameter.",
-      "_linkedTypes": [
+      "_embeddedTypes": [
         "https://openminds.ebrains.eu/core/QuantitativeValue",
         "https://openminds.ebrains.eu/core/QuantitativeValueRange"
       ]
--- a/schemas/research/numericalParameter.schema.tpl.json
+++ b/schemas/research/numericalParameter.schema.tpl.json
@@ -12,7 +12,6 @@
     "value": {
       "type": "array",
       "minItems": 1,
-      "uniqueItems": true,
       "_instruction": "Add at least one quantitative value for this parameter.",
       "_embeddedTypes": [
         "https://openminds.ebrains.eu/core/QuantitativeValue",
--- a/schemas/research/numericalParameter.schema.tpl.json
+++ b/schemas/research/numericalParameter.schema.tpl.json
@@ -1,0 +1,23 @@
+{
+  "_type": "https://openminds.ebrains.eu/core/NumericalParameter",
+  "required": [
+    "name",
+    "value"
+  ],
+  "properties": {
+    "name": {
+      "type": "string",
+      "_instruction": "Enter a descriptive name for this numerical parameter."
+    },
+    "value": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add at least one quantitative value for this numerical parameter.",
+      "_linkedTypes": [
+        "https://openminds.ebrains.eu/core/QuantitativeValue",
+        "https://openminds.ebrains.eu/core/QuantitativeValueRange"
+      ]
+    }
+  }
+}
--- a/schemas/research/parameterSetting.schema.tpl.json
+++ b/schemas/research/parameterSetting.schema.tpl.json
@@ -26,6 +26,6 @@
         "https://openminds.ebrains.eu/core/NumericalParameter",
         "https://openminds.ebrains.eu/core/StringParameter"
       ]
-    },
+    }
   }
 }
--- a/schemas/research/parameterSetting.schema.tpl.json
+++ b/schemas/research/parameterSetting.schema.tpl.json
@@ -1,19 +1,14 @@
 {
   "_type": "https://openminds.ebrains.eu/core/ParameterSetting",
   "required": [
-    "description",
-    "name",
+    "context",
     "relevantFor",
-    "value"
+    "setting"
   ],
   "properties": {
-    "description": {
+    "context": {
       "type": "string",
-      "_instruction": "Enter a description of this parameter setting."
-    },
-    "name": {
-      "type": "string",
-      "_instruction": "Enter the name of this parameter setting."
+      "_instruction": "Enter the common context for the parameters in this setting."
     },
     "relevantFor": {
       "_instruction": "Add the technique or behavioral task where this parameter setting is used in.",
@@ -22,15 +17,15 @@
         "https://openminds.ebrains.eu/controlledTerms/Technique"
       ]
     },
-    "value": {
-      "type": ["number", "string"],
-      "_instruction": "Enter the value of this parameter setting."
-    },
-    "unit": {
-      "_instruction": "Add the unit of measurement used for the value of this parameter setting.",
+    "parameters": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add all numerical and string parameters that belong to this setting.",
       "_linkedTypes": [
-        "https://openminds.ebrains.eu/controlledTerms/UnitOfMeasurement"
+        "https://openminds.ebrains.eu/core/NumericalParameter",
+        "https://openminds.ebrains.eu/core/StringParameter"
       ]
-    }
+    },
   }
 }
--- a/schemas/research/parameterSetting.schema.tpl.json
+++ b/schemas/research/parameterSetting.schema.tpl.json
@@ -3,7 +3,7 @@
   "required": [
     "context",
     "relevantFor",
-    "parameters"
+    "parameter"
   ],
   "properties": {
     "context": {
@@ -17,7 +17,7 @@
         "https://openminds.ebrains.eu/controlledTerms/Technique"
       ]
     },
-    "parameters": {
+    "parameter": {
       "type": "array",
       "minItems": 1,
       "uniqueItems": true,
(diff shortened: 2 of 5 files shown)


lzehl:
@apdavison let me know if you are too busy to review this, I can also ask Ulrike or Oli. Hope the changes will improve the logic a bit for the settings.

schemas/research/numericalParameter.schema.tpl.json
```diff
@@ -0,0 +1,23 @@
...
+    "value": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add at least one quantitative value for this numerical parameter.",
+      "_linkedTypes": [
+        "https://openminds.ebrains.eu/core/QuantitativeValue",
+        "https://openminds.ebrains.eu/core/QuantitativeValueRange"
+      ]
+    }
```
UlrikeS91: Why would you expect 1 - N `"value"` here? Can you give an example for when this would be useful? 

Personally, I can't come up with a good example. I think that name + value should be 1:1 (and you can still give a `quantitativeValueRange` if a single value is not enough). Additionally, `parameterSetting` allows the use of several numerical parameters. So, if you need several `"values"` for a `numericalParameter` related to the same `"name"`, one can just create several `numericalParameters` with the same `"name"` but different `"values"`, right?

schemas/research/parameterSetting.schema.tpl.json
```diff
@@ -1,19 +1,14 @@
 {
   "_type": "https://openminds.ebrains.eu/core/ParameterSetting",
   "required": [
-    "description",
-    "name",
+    "context",
     "relevantFor",
-    "value"
+    "setting"
```
UlrikeS91: Shouldn't this be `"parameter"`? The property `"setting"` at least doesn't exist. So, either `"parameter"` in line 20 needs to be changed to `"setting"` or vice versa.

schemas/research/parameterSetting.schema.tpl.json
```diff
@@ -22,14 +17,14 @@
...
     },
-    "value": {
-      "type": ["number", "string"],
-      "_instruction": "Enter the value of this parameter setting."
-    },
-    "unit": {
-      "_instruction": "Add the unit of measurement used for the value of this parameter setting.",
-      "_linkedTypes": [
-        "https://openminds.ebrains.eu/controlledTerms/UnitOfMeasurement"
+    "parameters": {
```
UlrikeS91: If you want to stick to `"parameters"` (see comment above), you should change the property name to its singular to stick to our logic that properties where one linkedType/embeddedType is a valid option, should be singular. 

--> change to `"parameter"` (or `"setting"`)

UlrikeS91 (changes requested):
See comments. There is one error that needs to be fixed (in `parameterSetting`) for sure.

@apdavison, your feedback to my question about the multiple values for `numericalParameter` would be very useful. Your general review as requested by @lzehl would be even better, of course :)

schemas/research/numericalParameter.schema.tpl.json
apdavison: What about adding ArrayParameter, to explicitly allow arrays of numerical values, and make NumericalParameter single valued?

schemas/research/stringParameter.schema.tpl.json
```diff
@@ -0,0 +1,17 @@
...
+    "value"
+  ],
+  "properties": {
+    "name": {
+      "type": "string",
+      "_instruction": "Enter a descriptive name for this string parameter."
+    },
+    "value": {
+      "type": "string",
+      "_instruction": "Enter a semantic value for this string parameter."
```
apdavison: I don't understand why "semantic" is used here

schemas/research/numericalParameter.schema.tpl.json
lzehl: One use case I imagined (quantitative values are simplified): 
"name": "stimulusTriggerPoints", 
"value": [1s, 2s, 4s]

In such a case, I would find entering all of them with separate names a bit cumbersome.

schemas/research/parameterSetting.schema.tpl.json
lzehl: :see_no_evil: that's a stupid one from my side. needs to be "parameters". I'll change it.

schemas/research/parameterSetting.schema.tpl.json
lzehl: We need to rethink this rule. We anyway do not stick to it fully (there are much more cases where this is not followed starting with our syntax "_linkedTypes", etc...) But I'll change it for now.

apdavison (commented):
I think my original idea is that `NumericalParameter` and `StringParameter` would replace `ParameterSetting`, whereas here it seems that `ParameterSetting` has become a container. I think that's fine, although in that case the name `ParameterSet` would be better, I think.

lzehl:
> I think my original idea is that `NumericalParameter` and `StringParameter` would replace `ParameterSetting`, whereas here it seems that `ParameterSetting` has become a container. I think that's fine, although in that case the name `ParameterSet` would be better, I think.

@apdavison yes that is something we discussed in the issue in order to facilitate the registration of highly related parameters without having to make this relation clear in the name of the parameter (e.g., used agent, used concentration of that agent).

I'm happy to change the Schema name into "ParameterSet" and the corresponding property as well in ProtocolExecution.

schemas/research/stringParameter.schema.tpl.json
lzehl: @apdavison I was searching for alternative word for "string". Any suggestions? "text value" ?

For the comment above: I don't see the advantage to introduce a PropertArray, also because it is only useful for numerical values, would it not? and would just allow multiple quantitativeValues instead of one. We never differentiated between this in other schemas, if both single or array entries are allowed. What do you think?

apdavison: Sure, "Enter a text value for this parameter" or just "Enter a value for this parameter".

I agree about ArrayParameter. What could however be useful is QuantitativeValueArray, which would make it easier to enter arrays if the units could be specified only once (e.g. if you have an array of size 100, having to specify the units separately for each element in the array makes it very verbose).

lzehl: @apdavison :smile: I just thought about the same thing for the quantitativeValues. I agree that this would be useful, the only problem I see is how to handle uncertainties in a QuantitativeValueArray. Any suggestions?

apdavison: For the uncertainty, it should be possible to specify either a single pair of values (same uncertainty for all values in the array), or a 2 x N array of uncertainty values.

lzehl: @apdavison If you agree, let's get back to the QuantitativeValueArray in the next revision. You think the remaining issues are solved in this PR?

lzehl: @apdavison feel free to merge it then when you are happy

apdavison:
@UlrikeS91 were all your requested changes taken into account?

schemas/research/numericalParameter.schema.tpl.json
```diff
@@ -0,0 +1,23 @@
...
+    "value": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add at least one quantitative value for this parameter.",
+      "_linkedTypes": [
+        "https://openminds.ebrains.eu/core/QuantitativeValue",
+        "https://openminds.ebrains.eu/core/QuantitativeValueRange"
+      ]
+    }
```
UlrikeS91: I think it's fine to have the array here. @lzehl example is valid, and it won't hurt to have it as an array. 

But: `"_linkedTypes":` is wrong, isn't it? We don't want to recycle quantitativeValue(Range). This should be `"_embeddedTypes":`, or?
If `"_linkedTypes":` is correct, `uniqueItems` should be removed. Then you may as well recycle the exact same value ;) But I'm pretty sure that we don't want them since we never allowed this before (see e.g. `specimenState` schema).

schemas/research/protocolExecution.schema.tpl.json
```diff
@@ -38,13 +38,13 @@
         "https://openminds.ebrains.eu/core/TissueSampleState"
       ]
     },
-    "parameterSetting": {
+    "parameterSet": {
       "type": "array",
       "minItems": 1,
       "uniqueItems": true,
       "_instruction": "Add all important parameter settings defining this protocol execution.",
```
UlrikeS91: Change to: "Add all important parameter sets defining this protocol execution."

(Does "sets" sound a bit strange, or is it just me?)

UlrikeS91 (changes requested):
@lzehl, I noticed some more things now. See comments.

UlrikeS91:
> @UlrikeS91 were all your requested changes taken into account?

@apdavison, yes, there was a lot of other things being discussed in parallel but I think everything was covered.

schemas/research/protocolExecution.schema.tpl.json
lzehl: @UlrikeS91 you are right in all comments. Thanks for being so thorough! I've updated the corresponding descriptions (mentioning setting) and the _embeddedTypes (incl. deletion of uniqueItems)

UlrikeS91 (approved):

Issue #190, opened 2021-03-21
Question: changes proposed to have a clearer usage of the term "instance"


lzehl:
@apdavison @olinux @UlrikeS91 @jagru20 @skoehnen @bweyers @Peyman-N

While writing the paper and the documentation on the future homepage for openMINDS (collab) and presenting the metadata models now several times I've frequently stumbled upon the term "instance". Here are the three aspects that are in connection with this term:

+ For describing information around a single file, we chose the schema label "FileInstance". 
+ We also have an openMINDS_instances repository that contains metadata representations of some schemas where we want the entries in the KG more controlled.
+ We have in openMINDS_controlledTerms a folder "terminologies" that contains metadata representations of the controlledTerms schemas (except for TermSuggestion).

I would like to make this more consistent and like to make the following proposal:
1) Change the schema name "FileInstance" to just "File"
2) Use the term "instance" when speaking about a metadata representation (JSON-LD) of a schema. That way the content of the openMINDS_instances repository is more clearly defined.
2) Move the instances of the "terminologies" in the openMINDS_controlledTerms repository to the openMINDS_instances repository.

What do you think?

UlrikeS91:
> 2. Use the term "instance" when speaking about a metadata representation (JSON-LD) of a schema. That way the content of the openMINDS_instances repository is more clearly defined.

Fine by me :) 

> 3. Move the instances of the "terminologies" in the openMINDS_controlledTerms repository to the openMINDS_instances repository.

Very good suggestion! Absolutely fine by me!

> 1. Change the schema name "FileInstance" to just "File"

Generally not opposed to changing the schema name. What was the reason to have "instance" in the name in the first place? Was it to have a clear solution (and difference) between "fileBundle", "fileRepository" and "file(Instance)"? I think "file" alone would be enough, just curious why it was added in the first place (and making sure that there wasn't any good reason for having it) ;)

lzehl:
@UlrikeS91 I actually do not remember :sweat_smile: But yes I think it was to have clear differentiation between the fileX schemas.
It came out of a discussion with Oli a very long time ago. @olinux do you remember why we decided for "FileInstance" instead of "File"?

lzehl:
Was solved.

PR #192, opened 2021-03-24
updating file schemas

lzehl:
related to #190 

Additionally:
@olinux split repo from bundle here, because one file can belong to multiple bundles but only one repository.


lzehl:
@olinux : I've also need to update SANDS considering the fileInstance->file change (@UlrikeS91 you could double check as well here)

lzehl:
this PR is also related to https://github.com/HumanBrainProject/openMINDS_core/issues/190

apdavison:
> one file can belong to multiple bundles but only one repository

can you explain this in more detail? Can the bundles be in different repositories?

If a file is copied from one repository to another, I would want to represent this as a single File object that is linked to both repositories. On that topic, are you in contact with anyone from T6.3? They are working on a data location service that I think should be based on File/FileBundle/FileRepository.

lzehl:
@apdavison based on a discussion with @olinux we agreed that a file bundle is defined by the files it is grouping. Within the KG a single file can only belong to one repository because we wanted to be able to assign clear/distinct download locations (full paths) for each file. A file that is copied to another repository would therefore get another entry. The relation of such copies are currently not represented in openMINDS. The file can link to multiple file bundles. If all the files that point to the same bundle are part of the same repository it can be inferred that this bundle is also part of this repository. In principle we are not forbidding that a file bundle could hold files of different repositories, although we try to avoid it by creating most file/file bundle relations automatically within the same file repository.

I was not yet with anyone of T6.3 I think. Is this connected to DataLad by any chance? Then indirectly in contact already.

apdavison:
The ability to link different copies of the same file is essential for provenance, and is probably also needed for the data location service. Perhaps we could add this issue to the agenda for when we meet to talk about openMINDS extensions?

lzehl:
@apdavison yes! I agree we should definitely put it on the agenda list. I made a note. Thanks for bringing it up.

UlrikeS91 (approved):

PR #196, opened 2021-03-30
make affiliation an embedded type

olinux:
... since it is tightly related to a person and doesn't have a lifecycle on its own


UlrikeS91:
@lzehl I'll take this one then since you took all mine from yesterday :P

UlrikeS91 (approved):

PR #199, opened 2021-04-14
Update softwareVersion.schema.tpl.json

jagru20:
changed "requirement" to "requirements" and string array
marked requirements as optional

### DIFF
--- a/schemas/products/softwareVersion.schema.tpl.json
+++ b/schemas/products/softwareVersion.schema.tpl.json
@@ -63,7 +63,7 @@
         "https://openminds.ebrains.eu/controlledTerms/SoftwareFeature"
       ]
     },
-    "requirements": {
+    "requirement": {
       "type": "array",
       "minItems": 0,
       "uniqueItems": true,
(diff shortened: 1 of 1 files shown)


schemas/products/softwareVersion.schema.tpl.json
```diff
@@ -64,9 +63,14 @@
         "https://openminds.ebrains.eu/controlledTerms/SoftwareFeature"
       ]
     },
-    "requirement": {
-      "type": "string",
-      "_instruction": "Enter all requirements of this software version."
+    "requirements": {
```
lzehl: as convention we decided to always use the singular for the property (even if it the expected value is an array

lzehl (changes requested):
just a minor correction: use the singular for the property despite the fact that the value can be an array

lzehl (approved):

lzehl:
@jagru20 there is one thing I overlooked which should be maybe changed is the "minItems". For a not required property it still should be 1 instead of 0, because it is not required (meaning you can just leave it out of the jsonld if you anyway would only have an empty array there). But I'd like to discuss this in the openMINDS development team again (I'll make an issue so we do not forget). Therefore I'm just merging your PR now.

PR #201, opened 2021-04-16
adding link to protocol to dataset version

lzehl:
this change will allow a very generic registration of a dataset version, which might get handy if more information is not available or if the additional information takes more time to gather.

### DIFF
--- a/schemas/products/datasetVersion.schema.tpl.json
+++ b/schemas/products/datasetVersion.schema.tpl.json
@@ -2,10 +2,11 @@
   "_type": "https://openminds.ebrains.eu/core/DatasetVersion",
   "_extends": "products/researchProductVersion.schema.tpl.json",
   "required": [
-    "ethicsAssessment",
     "digitalIdentifier",
-    "license",
+    "ethicsAssessment",
     "experimentalApproach",
+    "license",
+    "protocol",
     "type"
   ],
   "properties": {
@@ -28,6 +29,15 @@
       "_instruction": "Add the result of the ethics assessment of this dataset version.",
       "_linkedTypes": [
         "https://openminds.ebrains.eu/controlledTerms/EthicsAssessment"
+      ]
+    },
+    "experimentalApproach": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add all experimental approaches which this dataset version has deployed.",
+      "_linkedTypes": [
+        "https://openminds.ebrains.eu/controlledTerms/ExperimentalApproach"
       ]
     },
     "inputData": {
@@ -62,13 +72,13 @@
         "https://openminds.ebrains.eu/core/License"
       ]
     },
-    "experimentalApproach": {
+    "protocol": {
       "type": "array",
       "minItems": 1,
       "uniqueItems": true,
-      "_instruction": "Add all experimental approaches which this dataset version has deployed.",
+      "_instruction": "Add one or several protocols that were used in this dataset version.",
       "_linkedTypes": [
-        "https://openminds.ebrains.eu/controlledTerms/ExperimentalApproach"
+        "https://openminds.ebrains.eu/core/Protocol"
       ]
     },
     "studiedSpecimen": {
--- a/schemas/research/protocol.schema.tpl.json
+++ b/schemas/research/protocol.schema.tpl.json
@@ -23,11 +23,11 @@
       "type": "string",
       "_instruction": "Enter a descriptive name for this protocol."
     },
-    "studyTarget": {
+    "studyOption": {
       "type": "array",
       "minItems": 1,
       "uniqueItems": true,
-      "_instruction": "Add all study targets of this model version.",
+      "_instruction": "Add all study options of that this protocol offers.",
       "_linkedCategories": [
         "studyTarget"
       ]
--- a/schemas/research/protocol.schema.tpl.json
+++ b/schemas/research/protocol.schema.tpl.json
@@ -27,7 +27,7 @@
       "type": "array",
       "minItems": 1,
       "uniqueItems": true,
-      "_instruction": "Add all study options of that this protocol offers.",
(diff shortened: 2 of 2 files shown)


lzehl:
@olinux FYI that would require an update of the KG Editor

lzehl:
@UlrikeS91 if you have a better name for the "studyTarget" property in protocol (I just renamed it to "studyOption") let me know. Otherwise I was wondering if this property should be required then. What do you think?

UlrikeS91:
> @UlrikeS91 if you have a better name for the "studyTarget" property in protocol (I just renamed it to "studyOption") let me know. 

"studyOption" should do it for now. It's open enough to include the things we envisioned.  

> Otherwise I was wondering if this property should be required then. What do you think?

I would say no. We talked about the example with "technique" = "multi-electrode extracellular recording" where the additional information "studyOption" = "LFP" would be great to have. But another example could be technique = "primary antibody staining" + "secondary antibody staining", where I cannot see what could be added as "studyOption" with going towards the "studyTarget" (e.g. parvalbumin expressing cells) or the "parameterSet" (e.g. anit-PV antibody XY). "LFP" would not fit in either of these, right? It's a means to study your hypothesis and may has a specific set of parameters to measure it, but we wouldn't use the term itself for either of them.

schemas/products/datasetVersion.schema.tpl.json
```diff
@@ -2,10 +2,11 @@
   "_type": "https://openminds.ebrains.eu/core/DatasetVersion",
   "_extends": "products/researchProductVersion.schema.tpl.json",
   "required": [
-    "ethicsAssessment",
     "digitalIdentifier",
-    "license",
+    "ethicsAssessment",
     "experimentalApproach",
+    "license",
+    "protocol",
```
UlrikeS91, line 9: What is the idea behind making it a required property? If I have registered my protocols via protocolExecutions wouldn't it make more sense to leave the datasetVersion level blank then?

schemas/research/protocol.schema.tpl.json
```diff
@@ -23,11 +23,11 @@
       "type": "string",
       "_instruction": "Enter a descriptive name for this protocol."
     },
-    "studyTarget": {
+    "studyOption": {
       "type": "array",
       "minItems": 1,
       "uniqueItems": true,
-      "_instruction": "Add all study targets of this model version.",
+      "_instruction": "Add all study options of that this protocol offers.",
```
UlrikeS91: This sentence reads badly. Maybe: "Add all study options that this protocol offers."

UlrikeS91:
> > @UlrikeS91 if you have a better name for the "studyTarget" property in protocol (I just renamed it to "studyOption") let me know.
> 
> "studyOption" should do it for now. It's open enough to include the things we envisioned.
> 
> > Otherwise I was wondering if this property should be required then. What do you think?
> 
> I would say no. We talked about the example with "technique" = "multi-electrode extracellular recording" where the additional information "studyOption" = "LFP" would be great to have. But another example could be technique = "primary antibody staining" + "secondary antibody staining", where I cannot see what could be added as "studyOption" with going towards the "studyTarget" (e.g. parvalbumin expressing cells) or the "parameterSet" (e.g. anit-PV antibody XY). "LFP" would not fit in either of these, right? It's a means to study your hypothesis and may has a specific set of parameters to measure it, but we wouldn't use the term itself for either of them.

Thinking about this, maybe something around **"studyStrategy"** could also work? If we take "LFP" again, describing this as a strategy to study specific extracell. signals could work. Not sure if this would work for anything else...

I also quickly checked the PanProject list. "LFP" was listed under studyTarget but we may did this because there was no reason to find a better term/place to put it. Another word is stumbled over in the list was "aspect". Was this reserved for something specific? **"studyAspect" or "studiedAspect"** could be nice, right?

schemas/research/protocol.schema.tpl.json
lzehl: :joy: yes... my bad...

lzehl:
@UlrikeS91 both "strategy" and "aspects" are nice alternatives, but I would say more for the "studyTarget" as for this one here. My intuition would say that strategy/aspect and target fall in the same category ("I focused on that for some reason"). If it's alright with you let's keep studyOption now and maybe get back to Tom and Mathew about it.

schemas/products/datasetVersion.schema.tpl.json
lzehl: No. I want the double linkage then. I discussed that part also already with Oli. He is alright with it. For now the curators have to make the sanity check though that all protocols in use via protocolExecution are also attached to the DatasetVersion

lzehl: > > @UlrikeS91 if you have a better name for the "studyTarget" property in protocol (I just renamed it to "studyOption") let me know.
> 
> "studyOption" should do it for now. It's open enough to include the things we envisioned.
> 
> > Otherwise I was wondering if this property should be required then. What do you think?
> 
> I would say no. We talked about the example with "technique" = "multi-electrode extracellular recording" where the additional information "studyOption" = "LFP" would be great to have. But another example could be technique = "primary antibody staining" + "secondary antibody staining", where I cannot see what could be added as "studyOption" with going towards the "studyTarget" (e.g. parvalbumin expressing cells) or the "parameterSet" (e.g. anit-PV antibody XY). "LFP" would not fit in either of these, right? It's a means to study your hypothesis and may has a specific set of parameters to measure it, but we wouldn't use the term itself for either of them.

Completely agree. Oli as well. Let's leave it optional for the Protocol

UlrikeS91:
> @UlrikeS91 both "strategy" and "aspects" are nice alternatives, but I would say more for the "studyTarget" as for this one here. My intuition would say that strategy/aspect and target fall in the same category ("I focused on that for some reason"). If it's alright with you let's keep studyOption now and maybe get back to Tom and Mathew about it.

Absolutely!

UlrikeS91 (approved):

PR #202, opened 2021-04-19
Introduction of lookup labels for schemas with no human readable identifier.

lzehl:
As discussed this PR introduces optional lookup labels for schemas with no human readable identifier.


lzehl:
@UlrikeS91 there will be another PR for the same thing on openMINDS_SANDS

UlrikeS91 (approved):

PR #203, opened 2021-04-20
Introduction of a URL schema

lzehl:
Based on the discussions for "full documentation" we decided to introduce a schema for URLs


lzehl:
@olinux to be consistent I changed all homepages to link to a URL. That also includes Organizations where we would need to do an automatic migration within the KG.

There will be another PR in openMINDS_SANDS for the homepages there.

olinux (approved):

PR #204, opened 2021-04-21
extract activity schema

olinux:


schemas/research/protocolExecution.schema.tpl.json
```diff
@@ -1,56 +1,7 @@
 {
   "_type": "https://openminds.ebrains.eu/core/ProtocolExecution",
-  "required": [
-    "input", 
-    "output", 
-    "protocol"
```
lzehl, line 6: `protocol` still needs to be required in addition to `input` and `output` defined in `Activity`.

schemas/research/activity.schema.tpl.json
```diff
@@ -0,0 +1,71 @@
+{
+  "required": [
+    "input", 
+    "output"
+  ],
+  "properties": {
+    "description": {
+      "type": "string",
+      "_instruction": "Enter a description of this protocol execution."
```
lzehl: _instruction need to be generalized in the concept schema. (e.g., replace "protocol execution" with "activity")

schemas/research/activity.schema.tpl.json
```diff
@@ -0,0 +1,71 @@
...
+    },
+    "input": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add all inputs (subject state, tissue sample state, file instance or file bundle) used by this protocol execution.",
```
lzehl: _instruction need to be generalized in the concept schema. (e.g., replace "protocol execution" with "activity")
+ if I'm not mistaken, `_linkedTypes` will change for `input` in extensions, therefore the list in the brackets should be removed

schemas/research/activity.schema.tpl.json
```diff
@@ -0,0 +1,71 @@
...
+    },
+    "lookupLabel": {
+      "type": "string",
+      "_instruction": "Enter a lookup label for this protocol execution that may help you to more easily find it again."
+    },
+    "output": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add all outputs (subject state, tissue sample state, file instance or file bundle) generated by this protocol execution.",
```
lzehl: _instruction need to be generalized in the concept schema. (e.g., replace "protocol execution" with "activity")
+ if I'm not mistaken, `_linkedTypes` will change for `output` in extensions, therefore the list in the brackets should be removed

schemas/research/activity.schema.tpl.json
```diff
@@ -0,0 +1,71 @@
...
+        "https://openminds.ebrains.eu/core/File",
+        "https://openminds.ebrains.eu/core/SubjectGroupState",
+        "https://openminds.ebrains.eu/core/SubjectState",
+        "https://openminds.ebrains.eu/core/TissueSampleCollectionState",
+        "https://openminds.ebrains.eu/core/TissueSampleState"
+      ]
+    },
+    "lookupLabel": {
+      "type": "string",
+      "_instruction": "Enter a lookup label for this protocol execution that may help you to more easily find it again."
```
lzehl: _instruction need to be generalized in the concept schema. (e.g., replace "protocol execution" with "activity")

schemas/research/activity.schema.tpl.json
```diff
@@ -0,0 +1,71 @@
...
+        "https://openminds.ebrains.eu/core/SubjectState",
+        "https://openminds.ebrains.eu/core/TissueSampleCollectionState",
+        "https://openminds.ebrains.eu/core/TissueSampleState"
+      ]
+    },
+    "parameterSet": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add all important parameters grouped in context-specific sets that were used in this protocol execution.",
```
lzehl: _instruction need to be generalized in the concept schema. (e.g., replace "protocol execution" with "activity")

schemas/research/activity.schema.tpl.json
```diff
@@ -0,0 +1,71 @@
...
+      "_embeddedTypes": [
+         "https://openminds.ebrains.eu/core/ParameterSet"
+      ]
+    },
+    "semanticallyAnchoredTo": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add all anatomical entities to which the outputs of this protocol execution can be semantically anchored to.",
```
lzehl: _instruction need to be generalized in the concept schema. (e.g., replace "protocol execution" with "activity")

schemas/research/activity.schema.tpl.json
```diff
@@ -0,0 +1,71 @@
...
+      "_linkedTypes": [
+         "https://openminds.ebrains.eu/sands/AnatomicalEntity"
+      ]
+    },
+    "studyTarget": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add all study targets of this model version.",
```
lzehl: _instruction need to be generalized in the concept schema. (e.g., replace "protocol execution" with "activity")

lzehl (changes requested):
General comment: we need to decide if the "_instruction" should be generalized in a concept schema or if we want the "_instruction" to be always precise and therefore always be provided in the extension (actual schema) which would mean all properties need to be listed again... Or if we can avoid the "naming" of the actual schema in these instructions to make them precise for the property but generic for the schema. Let me know what you think and I'm happy to help implementing the decision we make.

lzehl:
@apdavison we need your help here. Should the Activity hold all the properties that are currently in there or should it only be "input" and "output" ??

schemas/research/activity.schema.tpl.json
```diff
@@ -0,0 +1,71 @@
...
+    },
+    "input": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add all inputs used by this activity.",
+      "_linkedTypes": [
+        "https://openminds.ebrains.eu/core/FileBundle",
+        "https://openminds.ebrains.eu/core/File",
+        "https://openminds.ebrains.eu/core/SubjectGroupState",
```
apdavison: I think `Activity` should have only `File` and `FileBundle` as input and output, the others should be added in the `input`/`output` fields of `ProtocolExecution`.

schemas/research/activity.schema.tpl.json
```diff
@@ -0,0 +1,71 @@
...
+        "https://openminds.ebrains.eu/core/SubjectState",
+        "https://openminds.ebrains.eu/core/TissueSampleCollectionState",
+        "https://openminds.ebrains.eu/core/TissueSampleState"
+      ]
+    },
+    "lookupLabel": {
```
apdavison, line 29: There is a widely used term for "lookup label": "tag" !

schemas/research/activity.schema.tpl.json
```diff
@@ -0,0 +1,71 @@
...
+      "_instruction": "Add all important parameters grouped in context-specific sets that were used in this activity.",
+      "_embeddedTypes": [
+         "https://openminds.ebrains.eu/core/ParameterSet"
+      ]
+    },
+    "semanticallyAnchoredTo": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add all anatomical entities to which the outputs of this activity can be semantically anchored to.",
```
apdavison: two "to"s is too many :-) Either "to which...anchored." or "which...anchored to". I prefer the former.

apdavison (commented):
looks good in general, a couple of minor things, and a discussion needed about what to put in input/output

schemas/research/activity.schema.tpl.json
lzehl: I agree

schemas/research/activity.schema.tpl.json
lzehl: We introduced lookup label now in many schemas. I would not like to change it again, and at least in the way it is used in the KG it is not a "tag" (for me at least). It is only meant for curators (or users) if they set up a metadata graph for a research product to be able to quicker find the entities again (we actually did not want to introduce it at all, but working purely with the UUIDs turned out to be too cumbersome). Therefore the lookup labels will never be visible in the KG Search. A "tag" could be misunderstood as a general tag (like a keyword) that might also appear in the Search later.

schemas/research/activity.schema.tpl.json
lzehl: You are right... we'll gonna correct this.

lzehl:
> looks good in general, a couple of minor things, and a discussion needed about what to put in input/output

@apdavison just to be sure: so besides input, output, description and lookup label, activity should have also the properties: study target and semantically anchored to, as well as parameter set?

schemas/research/activity.schema.tpl.json
apdavison: so it's more like an alias? (i.e. unique)

apdavison:
> > looks good in general, a couple of minor things, and a discussion needed about what to put in input/output
> 
> @apdavison just to be sure: so besides input, output, description and lookup label, activity should have also the properties: study target and semantically anchored to, as well as parameter set?

yes, I think so.

schemas/research/activity.schema.tpl.json
lzehl: it does not have to be unique, but it most likely will be (at least as far as a curator can think of a unique term for the whole KG system)

apdavison:
What do you think about adding fields "startedAtTime", "endedAtTime"? We add these in Computation, and for the ephys extension we would like to add them to subclasses of ProtocolExecution, so maybe having them in Activity, as the base class of all these, is the best place.

lzehl:
> What do you think about adding fields "startedAtTime", "endedAtTime"? We add these in Computation, and for the ephys extension we would like to add them to subclasses of ProtocolExecution, so maybe having them in Activity, as the base class of all these, is the best place.

@apdavison I don't have any objections having them in the activity base class. They should remain optional though, because in many cases these metadata will not be available for a protocol execution.

lzehl:
@apdavison just to be sure: we said now we'll remove the semantically anchored to from the ProtocolExecution / Activity schema correct? And reintroduce it in the TissueSample(Collection) schemas? And in the files?? I think this we need to discuss again (where to reintroduce a user assessed anatomical location of their data). 

Summary of necessary corrections: 
Activity input / output should be reduced to File / FileBundle.
Activity semanticallyAnchoredTo should be removed.
Activity should get startTime / endTime properties (with time and datetime as formats).

lzehl (approved):

lzehl:
@apdavison I'm merging this PR now, but it will cause some changes to be made in openMINDS_computation. I will go through them and either make a PR there or raise an issue as soon as I can.

PR #205, opened 2021-05-10
Allow 1-N protocols for a protocolExecution

UlrikeS91:
As discussed with @lzehl 
advantages:
- mokes the model more flexible 
- enables the creation of more generic protocols that can be mix-and-matched with specifics in the execution itself 
- reduces input/output dummies (e.g. tissue samples that the researcher did not have any information about but had to add it to create a protocol execution


lzehl (approved):

UlrikeS91:
@olinux FYI

PR #207, opened 2021-05-26
Update person schema

UlrikeS91:
It doesn't really make sense to have an array for the digitalIdentifier, now that only ORCIDs are allowed. A person should not have more than one ORCID. Right?

### DIFF
--- a/schemas/actors/person.schema.tpl.json
+++ b/schemas/actors/person.schema.tpl.json
@@ -8,10 +8,7 @@
   ],
   "properties": {
     "digitalIdentifier": {
-      "type": "array",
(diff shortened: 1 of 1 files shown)


jcolomb:
what identifier are allowed may change in the future,  isn't it ? 

having an array would allow to change what is allowed without changing the schema.

see https://github.com/jam-schema/jams/issues/20 for a similar discussion in a different context.

lzehl:
@UlrikeS91 : @jcolomb is fully correct, we currently only have ORCID in openMINDS, but there are more than one digitial identifiers for persons out there. It should stay therefore an array.

PR #209, opened 2021-06-15
Peyman changing semantically anchored to anatomical location

Peyman-N:
As the discussion between me, Andrew and Lyuba. I am moving the anatomical location from `protocolExecution` to the `tissueSample`. This was necessary as a given `protocolExecution` can act on multiple brain regions, therefore the relationship between `tissueSample` and its location becomes ambiguous. 
To solve this ambiguity as we discuss it, I have removed the `semanticallyAnchoredTo` property from `protocolExecution` and, added a new property named `anatomicalLocation` to the `tissueSample`.
This change would make both `protocolExecution` and `tissueSample` more intuitive. As currently it exist two methods for describing the target of a `protocolExecution`, one through `studyTarget` property and other one through `semanticallyAnchoredTo` property. While we there is no tool for giving any information about location of a `tissueSample`.


Peyman-N:
On a second note I wanted to make the `anatomicalLocation`'s type only `controlledTerms/UBERONParcellation` but this was not a decision that I could have made alone.

lzehl (commented):
@olinux seems good to me (except for the missing type), only question: should this PR be merged with yours (PR #204)?

schemas/research/tissueSample.schema.tpl.json
```diff
@@ -42,11 +42,21 @@
...
       ]
     },
+    "anatomicalLocation": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add all anatomical entities to which this tissue sample belongs.",
+      "_linkedTypes": [
+        "https://openminds.ebrains.eu/controlledTerms/UBERONParcellation",
+        "https://openminds.ebrains.eu/sands/AnatomicalEntity"
```
lzehl: Should also link to https://openminds.ebrains.eu/sands/ParcellationEntity

lzehl (changes requested):

lzehl:
> On a second note I wanted to make the `anatomicalLocation`'s type only `controlledTerms/UBERONParcellation` but this was not a decision that I could have made alone.

This should not be the case. UBERON is super generic and if you can be more precise you should be (either by providing a link to an atlas parcellation (parcellation entity) or to your own parcellation name (anatomical entity)).

schemas/research/tissueSample.schema.tpl.json
lzehl: And it should be https://openminds.ebrains.eu/sands/CustomAnatomicalEntity

Peyman-N: Done!

lzehl (approved):

lzehl:
@olinux if this does not collide with your PR #204 and you double checked the changes feel free to merge.

lzehl:
@Peyman-N I've just merged a PR from Oli (#204) which will require some conflict solving in your PR. It should be not too much trouble though. Could you have a look and resolve the conflicts?

lzehl:
@Peyman-N I've solved it already :smile: I've realized that there was an error in the master so I've corrected it directly while solving the conflict.

lzehl:
@olinux Activity Schema is now part of openMINDS_core,  the property "semanticallyAnchoredTo" is removed from the ProtocolExecution Schema, and the TissueSample Schema has a "anatomicalLocation" property now.

Peyman-N:
@lzehl sorry, I just saw your comment. :sweat_smile:  Anyhow thanks a lot :smile:

Issue #210, opened 2021-06-15
Reassessment of BehavioralTask


lzehl:
Currently, the BehavioralTask are assumed to be ControlledTerms. Many of the behavioral tasks we are getting, are though not well defined and/or are not officially registered as a meaningful task. 

I would like to make this a bit more user-oriented and flexible by introducing a user-defined BehavioralTask schema with required properties "Full name", "Short name", "Description", + optional a selection of properties that link to controlled terms (with instances) for the used stimulus, task components, task structure, task rules, task cognitive system, motor readout, etc. (to be discussed).

If we reassess the registration of behavioral tasks this way, we can immediately take over most tasks of the old system as they were registered (and partially described) there and over time register the respective controlled terms (for at least some task aspects) to increase the knowledge integration between tasks.

Here some example instances for the possible controlled terms for a behavioral task (@tgbugs feel free to make more/better suggestions):

STIMULUS
- visual cue
  - natural image
  - static grid
  - moving grid 
  - ...
- auditory cue
  - cosine-squared tone
  - one syllable word
  - ...
- tactile cue
  - ...

TASK STRUCTURE
- go/no-go
- saccade/anti-saccade 
- instructed delayed motor response
- center out
- ...

TASK COGNITIVE SYSTEM
- attention
- working-memory
- ...

MOTOR READOUT
- tap left finger
- tap right finger
- squeeze left toes
- squeeze right toes
- reach and grasp

@olinux , @apdavison , @Peyman-N , @UlrikeS91 , @visakhmr , @MBAbrams please correct and comment.
What do you think? Please make suggestions also in respect to Tier-3 (so that we do not collide, but built on top of each other).

jcolomb:
I hope this will not state the obvious. To my knowledge (which is restricted), attempts at ontologies for behavioural science failed. I was in a workshop in 2014 or 2015 (organised by Melissa Haendel ) and we realised that we could not agree on definitions, because much of it was dependent on the researcher's interpretation (that sometimes changes over time). 

In addition, description of behaviour seem to be very model system dependent, with human behaviour research being another world completely. For instance "motor readout" is very pregnant of an input-output view of behavior and behavior science, where others see and look for output-input problematics (actions). Also different motor outputs can get the same readout (different ways to press a lever for instance), the same assay could be used to test different things, etc

There is an issue between getting no information and getting a detailed information that the community would not agree on.


[...]

PR #212, opened 2021-06-23
Reassessment of BehavioralTask

lzehl:
Cf. #210 

1) Intro of BehavioralTask schema
For now only contains properties "fullName", "shortName", & "description".
Future: will be extended by properties linking to controlled terms that will characterize the behavioral task (in a more standardized way).

2) Remove "behavioralTask" property from Protocol schema. Protocols should be generic, but meaningful in order to be reusable across dataset versions (if possible). The BehavioralTask will be very use-case specific in most cases. For this reason its linkage should move to the ProtocolExecution.

3) Link BehavioralTask schema to ProtocolExecution schema (cf. reasons in point 2).

4) Link BehavioralTask schema to DatasetVersion schema. As for Protocols, it should be possible to link a BehavioralTask directly to the DatasetVersion without defining ProtocolExecutions. Reason: the provision of ProtocolExecutions takes time, because they define the unique linkage between input and output in a DatasetVersion. To be able to define an openMINDS "minimal" representation of what happened in a DatasetVersion without defining ProtocolExecutions, the shortcut linkage between Protocol / BehavioralTask and DatasetVersion was established.

### DIFF
--- a/schemas/products/datasetVersion.schema.tpl.json
+++ b/schemas/products/datasetVersion.schema.tpl.json
@@ -6,7 +6,7 @@
     "ethicsAssessment",
     "experimentalApproach",
     "license",
-    "protocol",
+    "technique",
     "type"
   ],
   "properties": {
@@ -81,15 +81,6 @@
         "https://openminds.ebrains.eu/core/License"
       ]
     },
-    "protocol": {
-      "type": "array",
-      "minItems": 1,
-      "uniqueItems": true,
-      "_instruction": "Add one or several protocols that were used in this dataset version.",
-      "_linkedTypes": [
-        "https://openminds.ebrains.eu/core/Protocol"
-      ]
-    },
     "studiedSpecimen": {
       "type": "array",
       "minItems": 1,
@@ -100,6 +91,15 @@
         "https://openminds.ebrains.eu/core/SubjectGroup",
         "https://openminds.ebrains.eu/core/TissueSample",
         "https://openminds.ebrains.eu/core/TissueSampleCollection"
+      ]
+    },
+    "technique": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add one or several techniques that were used in this dataset version.",
+      "_linkedTypes": [
+        "https://openminds.ebrains.eu/controlledTerms/Technique"
       ]
     },
     "type": {
--- a/schemas/products/datasetVersion.schema.tpl.json
+++ b/schemas/products/datasetVersion.schema.tpl.json
@@ -19,6 +19,15 @@
         "legalPerson"
       ]
     },
+    "behavioralTask": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add one or several behavioral tasks that were performed in this dataset version.",
+      "_linkedTypes": [
+        "https://openminds.ebrains.eu/core/BehavioralTask"
+      ]
+    },
     "digitalIdentifier": {
       "_instruction": "Add the globally unique and persistent digital identifier of this research product version.",
       "_linkedTypes": [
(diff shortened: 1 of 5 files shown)


UlrikeS91 (changes requested):
General first: The draft looks correct. Technically, it could be merged. But controlledTerms/BehavioralTask (+ instances) should be removed. This is (obviously) not included in this PR, but should be done in parallel. 

I didn't give any comment in the related issue, but the result of that discussion makes sense to me and how you intend to solve it here. 
There is only one thing that I'm wondering about: This solution puts BehavioralTasks and Protocols more or less on the same level. But in this PR, you added them with their own properties to DatasetVersion (DSV) and ProtocolExecution (PrEx). Wouldn't it make more sense to collect them under the same property? 
Specifically, I'm struggling with Protocol being a required property while BehavioralTask isn't. Does that mean that a BehavioralTask cannot replace a Protocol? Does it always need to be at least one Protocol and optionally BehavioralTask(s) in addition? I think I would have viewed a BehavioralTask as something similar to a Protocol and potentially something that can stand on its own (without a protocol). And yes, this is indirectly the answer I never gave to your last question in the issue #210 ;) 
Since both properties allow N Protocols/BehavioralTasks, it would not be a problem to merge it under one property name, unless you actively intended to require at least one protocol of course.

lzehl:
@UlrikeS91 thanks for your feedback. I actually thought about merging them into one property as you discuss above, but there is indeed a difference: a behavioral task although in principle could be done on its own, within an experiment it cannot be performed without having a technical protocol in place to record / image the behavior. The other way around: a technical protocol has to be used for any experiment, but does not necessarily require a behavioral task present as well. This is why we kept behavioral task separate from techniques in the past, and why it is not good to merge them (since we cannot state: should have at least one protocol, can have additionally more protocols or behavioral tasks). Does this answer your concern?

UlrikeS91:
> @UlrikeS91 thanks for your feedback. I actually thought about merging them into one property as you discuss above, but there is indeed a difference: a behavioral task although in principle could be done on its own, within an experiment it cannot be performed without having a technical protocol in place to record / image the behavior. The other way around: a technical protocol has to be used for any experiment, but does not necessarily require a behavioral task present as well. This is why we kept behavioral task separate from techniques in the past, and why it is not good to merge them (since we cannot state: should have at least one protocol, can have additionally more protocols or behavioral tasks). Does this answer your concern?

Yes, this makes sense :) Then it also makes sense to keep them separate and to keep BehavioralTask optional (0 - N), while Protocol is required (1 - N). 

I think this can change from draft to reality now ;)

UlrikeS91 (approved):

lzehl:
@Peyman-N & @apdavison this PR might affect you as well.

UlrikeS91:
I just noticed that ParameterSet needs some adjustments because of this, too, @lzehl.

lzehl:
@UlrikeS91 why?

UlrikeS91:
Because it has behavioralTask as well: 
````
    "relevantFor": {
      "_instruction": "Add the technique or behavioral task where this set of parameters is used in.",
      "_linkedTypes": [
        "https://openminds.ebrains.eu/controlledTerms/BehavioralTask",
        "https://openminds.ebrains.eu/controlledTerms/Technique"
      ]
    }
````
Shouldn't that at least point to the new schema for BehavioralTask then?

lzehl:
@UlrikeS91 mh... not sure... let's do it though and at least it needs to change the type.

lzehl:
@UlrikeS91 done

lzehl:
@UlrikeS91 @Peyman-N & @apdavison FYI

Based on requests from Jan to simplify the minimal openMINDS metadata model even more and based on issues that came up during (partially respective) technical discussions with Oli the following changes will be made in this PR as well.

1) Dataset Version will not link to Protocols but to Techniques. [minimal track request from Jan]
2) Protocol Executions will link additionally to Dataset Version (required; count 1) to emphasize the fact that a PE cannot be reused across Dataset Versions. [idea from @olinux; will require later on a graph validation]

@apdavison if the later point is also true for all your Activity extending schemas we could later move this linkage also to Activity.

olinux (approved):

lzehl:
@olinux done. you can verify and merge if everything is in place.

PR #216, opened 2021-07-30
Adding mediatypes

jagru20:
@apdavison could you or your team have a spot-check on these mediatypes? Just to make sure there are no systematic errors in the jsonlds. Thanks!


apdavison (changes requested):
There is some redundancy here with respect to existing entries. One I spotted is that igor.jsonld is essentially identical to the existing igorpro.jsonld.

Can I suggest using the file extensions to identify other possible duplicates?

apdavison:
Possibly a bigger issue: shouldn't this PR be made against https://github.com/HumanBrainProject/openMINDS_instances ?

Or has the process been changed? @lzehl can you comment?

lzehl:
@apdavison there was an update on where to store the instances (announced via slack). Sorry if that got lost. I think I should do such announcements via email... We moved the instances to the corresponding metadata model repos (again) because they are version dependent, and we decided it is easier to handle them then directly within the corresponding version branch. If they are backwards compatible we should have them on all version branches of course (this might not yet be the case, I'm still organizing the cleanup). Another question though: how should we handle the openMINDS_instances GitHub? Should we delete it after making an announcement on the README or should we keep it?

lzehl:
@jagru20 & @apdavison for the potential duplicates: the extensions might not help very much to identify duplicates. On the one hand I expect that two versions of the "same format" will get separate entries, but would most likely have the same extension(s). On the other hand all the ambiguous extensions (e.g., csv, tsv, json, etc) which have one generic media type, but can have multiple content types for specific applications will pop up as well.

lzehl:
@jagru20 & @apdavison another question: what about the generic media types (csv, tsv, json, etc)? Should the "relatedMediaType" link to those also for the application specific content type? Or should we allow multiple content types for files, file bundles, and file repositories, so that they can link to the generic & the specific content type?

lzehl (changes requested):
I agree with @apdavison that the potential duplicates should be identified and eliminated. 

I would suggest to add the matching "relatedMediaType" to the entries which definitely have one (e.g. image/vnd.adobe.photoshop -> https://www.iana.org/assignments/media-types/image/vnd.adobe.photoshop)

apdavison:
Where we have application-specific content types that are based on a more generic type, the generic type is generally indicated with a "+".

For example, for IPython/Jupyter notebooks the content type is "application/x-ipynb+json", so indicating that any application that can read JSON can open this file, although of course the full functionality is only available if the application can interpret the JSON as a Jupyter notebook.

lzehl:
@apdavison I know, but should we still use a direct connection (either via "relatedMediaType" or via multiple contentTypes) to indicate that all CSVs (independent of their specification) can be viewed with the tools registered for the generic CSV media type?

apdavison:
@lzehl I think all three approaches would work, i.e.:
1. direct connection via "relatedMediaType"
2. direct connection via multiple "contentType"s
3. parse the content type name to use the "+"part

I don't have a strong preference, I have a weak preference for option 3.

p.s. maybe we should move this discussion to a separate issue, I don't think this PR should wait on the outcome.

lzehl:
@apdavison I agree. We can adopt later. But as a comment: I think 3 should be always used (based on the naming convention), and in addition we can introduce 1 or 2 (or 4: intro of property "parentMediaType").

lzehl:
@jagru20 do you need help for this PR ?

jagru20:
@lzehl sorry for the delay in this pr. Of course do the redundancies need to be eliminated. I think a little help would be quite useful here. I hope I understand the above discussion correctly when assuming that we add any relatey mediatypes at a later point in time?

instances/data/contentTypes/igor.jsonld
```diff
@@ -0,0 +1,16 @@
+{
```
jagru20: can be deleted, because redundant

instances/data/contentTypes/moleculardevices.axon.jsonld
```diff
@@ -0,0 +1,15 @@
+{
```
jagru20, line 1: this is redundant to [abf.jsonld](https://github.com/HumanBrainProject/openMINDS_core/blob/v3/instances/data/contentTypes/abf.jsonld) But I would prefer the file naming of this version, as it is more precise, what do you think @apdavison @lzehl ?

instances/data/contentTypes/moleculardevices.axon.jsonld
```diff
@@ -0,0 +1,15 @@
...
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.moleculardevices.axon",
+	"@type": "https://openminds.ebrains.eu/core/ContentType",
+	"fileExtension": [
+		".abf"
+	],
+	"name": "application/vnd.moleculardevices.axon",
+	"relatedMediaType": null,
+	"synonym": [
+		"Axon™ Binary File format"
```
jagru20, line 13: is the use of the (TM) character a problem?

lzehl: @olinux should we support markdown for all text fields? this formatting is not only an issue/question for this item but all text fields within openMINDS.

instances/data/contentTypes/moleculardevices.axon.jsonld
lzehl: I leave this to @apdavison

lzehl:
> @lzehl sorry for the delay in this pr. Of course do the redundancies need to be eliminated. I think a little help would be quite useful here. I hope I understand the above discussion correctly when assuming that we add any related media types at a later point in time?

@jagru20 yes the media types can be added later.

instances/data/contentTypes/neo.ascii.image.jsonld
```diff
@@ -0,0 +1,15 @@
+{
```
jagru20, line 1: this one is redundant to [asciimage.jsonld](https://github.com/HumanBrainProject/openMINDS_core/blob/v3/instances/data/contentTypes/asciiimage.jsonld) but I prefer this filename, as it seems more precise to me what do you think @apdavison ? You could also just let me know in general, that should prevent your inbox from being full of these requests.

instances/data/contentTypes/neo.ascii.signal.jsonld
```diff
@@ -0,0 +1,18 @@
+{
```
jagru20, line 1: redundant https://github.com/HumanBrainProject/openMINDS_core/blob/v3/instances/data/contentTypes/asciisignal.jsonld

instances/data/contentTypes/neo.ascii.spiketrain.jsonld
```diff
@@ -0,0 +1,15 @@
+{
```
jagru20, line 1: redundant https://github.com/HumanBrainProject/openMINDS_core/blob/v3/instances/data/contentTypes/asciispiketrain.jsonld

lzehl:
@jagru20 for the redundant entries: should we identify them within the review of this PR? this might be a bit cumbersome... in any case I maybe able to help you towards the end of this week.

jagru20:
I have prepared a commit, where I tried to eliminate the redundancies and keep as much as information as possible available.  Although I would prefer to hear @apdavison thoughts on the filename thing first, I can also just push it and we discuss that later as the essential information in the file contents is the same or extended.

What do you think is the best way to go here?

instances/data/contentTypes/neo.ascii.image.jsonld
apdavison: @jagru20 I'm happy for you to modify the filenames

instances/data/contentTypes/moleculardevices.axon.jsonld
apdavison: This is not a Markdown question, I don't think. I think the text fields should be unicode, UTF-8 encoding. ™ is therefore a valid character.

lzehl: yes, Andrew is right. it should be fine

instances/data/contentTypes/blackrockmicrosystems.cerebus.jsonld
```diff
@@ -0,0 +1,15 @@
+{
+	"@context" : {
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.blackrockmicrosystems.cerebus",
+	"@type": "https://openminds.ebrains.eu/core/ContentType",
+	"fileExtension": [
+		".nsx"
```
lzehl: I think this needs to be changed to the specific extensions (.ns1, .ns2, .ns3, .ns4, .ns5, .ns6)

There is also a neural event data format from blackrock (.nev). If not captured yet this should be added as a content type as well

instances/data/contentTypes/cml.jsonld
```diff
@@ -0,0 +1,13 @@
+{
+	"@context" : {
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/chemical/x-cml",
```
lzehl: what does "chemical" mean in this context??

instances/data/contentTypes/fmri.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+	"@context" : {
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.fmri",
```
lzehl: I don't get this one... if this is .nii OR .nii.gz means this is NIftI (format 1 or 2?) in dimensions (2D, 3D, 4D) the latter dimension file contains time data as well (but I do not know that this has the extension .dtseries.nii).
What is software connected to this?

instances/data/contentTypes/ge-healthcare-life-sciences.amersham-biosciences-gel.jsonld
```diff
@@ -0,0 +1,13 @@
...
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.ge-healthcare-life-sciences.amersham-biosciences-gel",
+	"@type": "https://openminds.ebrains.eu/core/ContentType",
+	"fileExtension": [
+		".gel"
+	],
+	"name": "application/vnd.ge-healthcare-life-sciences.amersham-biosciences-gel",
+	"relatedMediaType": null,
+	"synonym": null
```
lzehl: is there really no better synonym for this quite long and complicated name?

instances/data/contentTypes/hyland.brainwaredam.jsonld
```diff
@@ -0,0 +1,15 @@
+{
+	"@context" : {
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.hyland.brainwaredam",
```
lzehl, line 5: I guess brainwaredam.jsonld is a duplicate and needs to be deleted then??

instances/data/contentTypes/hyland.brainwaref32.jsonld
```diff
@@ -0,0 +1,15 @@
+{
+	"@context" : {
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.hyland.brainwaref32",
```
lzehl, line 5: brainwaref32.jsonld is then a duplicate and needs to be deleted?

instances/data/contentTypes/hyland.brainwaresrc.jsonld
```diff
@@ -0,0 +1,15 @@
+{
+	"@context" : {
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.hyland.brainwaresrc",
```
lzehl, line 5: brainwaresrc.jsonld is a duplicate then and needs to be deleted?

instances/data/contentTypes/lavision.imspector.jsonld
```diff
@@ -0,0 +1,15 @@
+{
+	"@context" : {
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.lavision.imspector",
```
lzehl, line 5: is this the same as imspector.jsonld ??

instances/data/contentTypes/matlab.jsonld
```diff
@@ -1,15 +1,15 @@
...
-  "fileExtension": [
-    ".m"
-  ],
-  "name": "text/x-matlab",
-  "relatedMediaType": null,
-  "synonym": [
+	"@context" : {
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/text/x-matlab",
```
lzehl: filename should be x-matlab as well. Question: what about matlab files with .mat ? is this a different content type?

instances/data/contentTypes/minc.jsonld
```diff
@@ -0,0 +1,15 @@
+{
+	"@context" : {
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.minc",
```
lzehl, line 5: there are different minc versions as far as I know. do we cover them?

instances/data/contentTypes/nest.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+	"@context" : {
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.nest",
```
lzehl, line 5: what software is this?

instances/data/contentTypes/nifti.1.jsonld
```diff
@@ -0,0 +1,18 @@
+{
+	"@context" : {
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.nifti.1",
```
lzehl, line 5: could you update nifti.2 accordingly?

instances/data/contentTypes/ome.tiff.jsonld
```diff
@@ -0,0 +1,18 @@
+{
+	"@context" : {
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.ome.tiff",
+	"@type": "https://openminds.ebrains.eu/core/ContentType",
+	"fileExtension": [
+		".ome.btf",
+		".ome.tf8",
+		"ome.tf2",
```
lzehl: dot missing

instances/data/contentTypes/ome.tiff.jsonld
```diff
@@ -0,0 +1,18 @@
...
+		"ome.tiff"
```
lzehl: dot missing

instances/data/contentTypes/python.jsonld
```diff
@@ -0,0 +1,13 @@
+{
+	"@context" : {
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.python",
```
lzehl: do we need to cover the different versions?

instances/data/contentTypes/raw.binarysignal.jsonld
```diff
@@ -0,0 +1,15 @@
+{
+	"@context" : {
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.raw.binarysignal",
```
lzehl, line 5: is this different to raw.jsonld?

instances/data/contentTypes/raw.mcs.jsonld
```diff
@@ -0,0 +1,15 @@
+{
+	"@context" : {
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.raw.mcs",
```
lzehl, line 5: rawmcs.jsonld is a duplicate then and should be deleted?

instances/data/contentTypes/traces .jsonld
```diff
@@ -0,0 +1,13 @@
+{
+	"@context" : {
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.traces ",
```
lzehl: delete space

instances/data/contentTypes/traces .jsonld
```diff
@@ -0,0 +1,13 @@
...
+	"@type": "https://openminds.ebrains.eu/core/ContentType",
+	"fileExtension": [
+		".xml"
+	],
+	"name": "application/vnd.traces ",
```
lzehl: delete space

lzehl (changes requested):
Hi @jagru20 thanks a lot for this huge PR. Are all deleted files replaces by a more accurate one from you? I think I did not find a matching new one for all of the deleted once. 

@apdavison some of my comments might be also solvable by you.

lzehl:
@jagru20 one more thing I noticed: the blackrock formats need to be splitted I think into .ns2 - .ns6 and .nev...

jagru20:
> 
> 
> @jagru20 one more thing I noticed: the blackrock formats need to be splitted I think into .ns2 - .ns6 and .nev...

you mean splitting them into separate contenttypes instead of listing the separate extensions in the application/vnd.blackrockmicrosystems.cerebus?

lzehl:
> > @jagru20 one more thing I noticed: the blackrock formats need to be splitted I think into .ns2 - .ns6 and .nev...
> 
> you mean splitting them into separate contenttypes instead of listing the separate extensions in the application/vnd.blackrockmicrosystems.cerebus?

@jagru20 yes. At least nev vs nsX need to be different content types because they need different loading routines. For the nsX we can discuss. @apdavison what is your though about this ?

apdavison:
My understanding is that a given recording session is typically split over a nev and one or more nsX files with the same file name prefix, so you need all files to correctly load the data. This is how Neo handles them in any case: see https://github.com/NeuralEnsemble/python-neo/blob/master/neo/rawio/blackrockrawio.py

Probably we should have a separate content type for each, plus maybe a content type applied to a FileBundle that would represent the group of files belonging to the same recording.

lzehl:
@apdavison sounds good. One contentType for each and one for "blackrock parallel recordings". Thank you for the quick feedback :)

jagru20:
Sorry for double asking but "for each" in this case means one for all .nsx files, listing their possible extensions and one for .nev, right? Otherwise, this would not make sense because, the loading routine for different .nsx files would only differ depending on their header contents right?

instances/data/contentTypes/cml.jsonld
jagru20: this is the type naming as proposed by the chemical MIME project (cf. https://www.ch.ic.ac.uk/chemime/ or https://en.wikipedia.org/wiki/Chemical_Markup_Language) but as this is unofficial and "only" de facto standard since 1998 and apparently xml based, I think we could change this to  `text/x-cml+xml`

instances/data/contentTypes/fmri.jsonld
jagru20: I think the id/name should be reworked here. The fileformat is CIFTI which makes use of NIFTI-2 (see https://wiki.humanconnectome.org/display/PublicData/HCP+Users+FAQ) there it also states that these files must not be gzipped if you want to access it, but can be zipped for transferring them. I think this is why this gets confusing. I suggest splitting this up or deleting the .nii.gz extension, as this does not seem to add value here, it needs to be unzipped anyway most likely

lzehl:
> Sorry for double asking but "for each" in this case means one for all .nsx files, listing their possible extensions and one for .nev, right? Otherwise, this would not make sense because, the loading routine for different .nsx files would only differ depending on their header contents right?

@apdavison please correct me if I'm wrong: we would need contentTypes for each nsX and one for nev. If I remember correctly the nsX differ in their structure, because they are supported by different versions? @jagru20 sorry for the confusion, but it's too long ago that I worked with those files... In addition, Andrew suggested to have a blackrock contentType for parallel recordings (for a file bundle) that can indicate when several nsX and a nev are actually recorded in parallel and should be looked at together.

instances/data/contentTypes/cml.jsonld
lzehl: your suggestion sounds nicer (more consistent with the others) to me. @apdavison what do you think?

instances/data/contentTypes/fmri.jsonld
lzehl: Why not make use of the name "CIFTI-2" then? in combination with NIFTI-2 if that is the base? @apdavison your opinion/suggestion? (I don't want to suggest a name since the two of you are much more familiar with the convention)

jagru20: yes, that's what I meant. I suggest `application/fmri.cifti-2` note that I did not add a `+nifti` suffix because I am not sure if (and don't think) it is possible to open a cifti-2 file with any generic nifti-2 viewer. 

However, I would add the link to the specification file also: https://www.nitrc.org/forum/attachment.php?attachid=341&group_id=454&forum_id=1955

lzehl: looks good :+1:

instances/data/contentTypes/ge-healthcare-life-sciences.amersham-biosciences-gel.jsonld
jagru20: I'm adding "GEL" as a synonym.

instances/data/contentTypes/lavision.imspector.jsonld
jagru20: I am not completely sure about this, but I think they are different, because they are from different companies (abberior vs. lavision biotec) and bound to software distributed with their hardware. I would add abberior to the name of the other imspector to make the differnce clear.

instances/data/contentTypes/matlab.jsonld
jagru20: The "x-" prefix is usually (at IANA) used to mark that a contenttype is experimental, thus not fully specified. I rather suggest deleting the prefix. This also applies to several other contenttypes.

instances/data/contentTypes/minc.jsonld
jagru20: Not yet, because apparently there was no need to distinguish between them until now, I think.

instances/data/contentTypes/nest.jsonld
jagru20: nest is a simulator for spiking neural networks, see https://nest-simulator.org/

instances/data/contentTypes/python.jsonld
jagru20: I don't think so. If I am correct, the file specification of a python script file should not be affected by the version it makes use of, right?

instances/data/contentTypes/raw.binarysignal.jsonld
jagru20: no, that one is a panasonic specific raw image type

instances/data/contentTypes/python.jsonld
apdavison: We may at least wish to distinguish Python 2 and Python 3. Even within Python 3, later versions add new syntax, so a script written with Python 3.10 features will not run with  Python 3.7.

instances/data/contentTypes/blackrockmicrosystems.cerebus.jsonld
jagru20: I'll divide this up into a separate contenttype for each of the .nsx types named `application/vnd.blackrockmicrosystems.continuousdata.x` where x is replaced by the number in the respective extension. Moreover, I add the .nev contenttype as `blackrockmicrosystems.spikeinfo.jsonld`

lzehl: @jagru20 I think those would be better (because these are the full words of the abbreviations used for the extensions)
NSx: application/vnd.blackrockmicrosystems.neuroalsignals.x
NEV: application/vnd.blackrockmicrosystems.neuralevents

instances/data/contentTypes/python.jsonld
jagru20: ok, I see the point. I have no objections adding separate contenttypes for pythonfiles, I would do it in a manner of `application/vnd.python.2` or to distinguish further `application/vnd.python.23`, `application/vnd.python.31` etc. What do you think?

lzehl: @apdavison I do not know, what do you think?

instances/data/contentTypes/blackrockmicrosystems.cerebus.jsonld
jagru20: I agree. Is it `neuralsignals` or `neuronalsignals`? It's `neural` I suppose?

lzehl: @apdavison do you know ? would be the same for both `neural` or `neuronal` + `signals` OR `events` ??

jagru20: I had a quick look at the specification, it is neural https://blackrockneurotech.com/research/wp-content/ifu/LB-0023-7.00_NEV_File_Format.pdf c.f. pages 3 and 16 respectively

instances/data/contentTypes/python.jsonld
apdavison: There is an unofficial but widely used mediatype for Python code: `text/x-python`
I'd therefore go with `text/x-python`, `text/x-python.2`, `text/x-python.3`,  `text/x-python.2.7`, etc. I think it's better to keep the "." in the version identifiers.

jagru20: ok i'll then go with your suggestion.

jagru20: one problem that I have with the "x-" prefix though is that it is usually used to mark "experimental" contenttypes where "experimental" seems to mean "not IANA registered" would that mean, that we have to mark every not-IANA registered, type with that prefix? see also [matlab](https://github.com/HumanBrainProject/openMINDS_core/pull/216#discussion_r727060811)

apdavison: That's complicated. My understanding is that use of the "x-" prefix has been deprecated, but where there are well-established media types with this prefix they can be retained.

jagru20: Ok so we keep it where it is used by now but do not introduce it for future types.

instances/data/contentTypes/matlab.jsonld
jagru20: We'll proceed as discussed [here](https://github.com/HumanBrainProject/openMINDS_core/pull/216#discussion_r727772112)

instances/data/contentTypes/blackrockmicrosystems.neuralevents.jsonld
```diff
@@ -0,0 +1,16 @@
...
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.blackrockmicrosystems.neuralevents",
+	"@type": "https://openminds.ebrains.eu/core/ContentType",
+	"fileExtension": [
+		".nev"
+	],
+	"specification": "https://blackrockneurotech.com/research/wp-content/ifu/LB-0023-7.00_NEV_File_Format.pdf",
+	"name": "application/vnd.blackrockmicrosystems.neuralevents",
+	"relatedMediaType": null,
+	"synonym": [
+    "Cerebus neural signal system"
```
lzehl: I would add "Blackrock Neural Events"

instances/data/contentTypes/blackrockmicrosystems.neuralsignals.1.jsonld
```diff
@@ -0,0 +1,17 @@
...
+	"@type": "https://openminds.ebrains.eu/core/ContentType",
+	"fileExtension": [
+		".ns1"
+	],
+	"specification": "https://blackrockneurotech.com/research/wp-content/ifu/LB-0023-7.00_NEV_File_Format.pdf",
+	"name": "application/vnd.blackrockmicrosystems.neuralsignals",
+	"relatedMediaType": null,
+	"synonym": [
+    "Cerebus neural signal system",
+	"Blackrock Neural Signal 1"
```
lzehl: correct indent, change to "Blackrock Neural Signal**s** 1"

instances/data/contentTypes/blackrockmicrosystems.neuralsignals.2.jsonld
```diff
@@ -0,0 +1,17 @@
...
+	"@type": "https://openminds.ebrains.eu/core/ContentType",
+	"fileExtension": [
+		".ns2"
+	],
+	"specification": "https://blackrockneurotech.com/research/wp-content/ifu/LB-0023-7.00_NEV_File_Format.pdf",
+	"name": "application/vnd.blackrockmicrosystems.neuralsignals",
+	"relatedMediaType": null,
+	"synonym": [
+    "Cerebus neural signal system",
+	"Blackrock Neural Signal 2"
```
lzehl: correct indent, change to "Blackrock Neural Signals 2"

instances/data/contentTypes/blackrockmicrosystems.neuralsignals.3.jsonld
```diff
@@ -0,0 +1,17 @@
...
+	"@type": "https://openminds.ebrains.eu/core/ContentType",
+	"fileExtension": [
+		".ns3"
+	],
+	"specification": "https://blackrockneurotech.com/research/wp-content/ifu/LB-0023-7.00_NEV_File_Format.pdf",
+	"name": "application/vnd.blackrockmicrosystems.neuralsignals",
+	"relatedMediaType": null,
+	"synonym": [
+    "Cerebus neural signal system",
+	"Blackrock Neural Signal 3"
```
lzehl: correct indent, change to "Blackrock Neural Signals 3"

instances/data/contentTypes/blackrockmicrosystems.neuralsignals.4.jsonld
```diff
@@ -0,0 +1,17 @@
...
+	"@type": "https://openminds.ebrains.eu/core/ContentType",
+	"fileExtension": [
+		".ns4"
+	],
+	"specification": "https://blackrockneurotech.com/research/wp-content/ifu/LB-0023-7.00_NEV_File_Format.pdf",
+	"name": "application/vnd.blackrockmicrosystems.neuralsignals",
+	"relatedMediaType": null,
+	"synonym": [
+    "Cerebus neural signal system",
+	"Blackrock Neural Signal 4"
```
lzehl: correct indent, change to "Blackrock Neural Signals 4"

instances/data/contentTypes/blackrockmicrosystems.neuralsignals.5.jsonld
```diff
@@ -0,0 +1,17 @@
...
+	"@type": "https://openminds.ebrains.eu/core/ContentType",
+	"fileExtension": [
+		".ns5"
+	],
+	"specification": "https://blackrockneurotech.com/research/wp-content/ifu/LB-0023-7.00_NEV_File_Format.pdf",
+	"name": "application/vnd.blackrockmicrosystems.neuralsignals",
+	"relatedMediaType": null,
+	"synonym": [
+    "Cerebus neural signal system",
+	"Blackrock Neural Signal 5"
```
lzehl: correct indent, change to "Blackrock Neural Signals 5"

instances/data/contentTypes/blackrockmicrosystems.neuralsignals.6.jsonld
```diff
@@ -0,0 +1,17 @@
...
+	"@type": "https://openminds.ebrains.eu/core/ContentType",
+	"fileExtension": [
+		".ns6"
+	],
+	"specification": "https://blackrockneurotech.com/research/wp-content/ifu/LB-0023-7.00_NEV_File_Format.pdf",
+	"name": "application/vnd.blackrockmicrosystems.neuralsignals",
+	"relatedMediaType": null,
+	"synonym": [
+    "Cerebus neural signal system",
+	"Blackrock Neural Signal 6"
```
lzehl: correct indent, change to "Blackrock Neural Signals 6"

instances/data/contentTypes/blackrockmicrosystems.neuralsignals.7.jsonld
```diff
@@ -0,0 +1,17 @@
...
+	"@type": "https://openminds.ebrains.eu/core/ContentType",
+	"fileExtension": [
+		".ns7"
+	],
+	"specification": "https://blackrockneurotech.com/research/wp-content/ifu/LB-0023-7.00_NEV_File_Format.pdf",
+	"name": "application/vnd.blackrockmicrosystems.neuralsignals",
+	"relatedMediaType": null,
+	"synonym": [
+    "Cerebus neural signal system",
+	"Blackrock Neural Signal 7"
```
lzehl: correct indent, change to "Blackrock Neural Signals 7"
I did not know that this existed... what is different here?

instances/data/contentTypes/blackrockmicrosystems.neuralsignals.8.jsonld
```diff
@@ -0,0 +1,17 @@
...
+	"@type": "https://openminds.ebrains.eu/core/ContentType",
+	"fileExtension": [
+		".ns8"
+	],
+	"specification": "https://blackrockneurotech.com/research/wp-content/ifu/LB-0023-7.00_NEV_File_Format.pdf",
+	"name": "application/vnd.blackrockmicrosystems.neuralsignals",
+	"relatedMediaType": null,
+	"synonym": [
+    "Cerebus neural signal system",
+	"Blackrock Neural Signal 8"
```
lzehl: correct indent, change to "Blackrock Neural Signals 8"
I did not know that this existed... what is different here?

instances/data/contentTypes/blackrockmicrosystems.neuralsignals.9.jsonld
```diff
@@ -0,0 +1,17 @@
...
+	"@type": "https://openminds.ebrains.eu/core/ContentType",
+	"fileExtension": [
+		".ns9"
+	],
+	"specification": "https://blackrockneurotech.com/research/wp-content/ifu/LB-0023-7.00_NEV_File_Format.pdf",
+	"name": "application/vnd.blackrockmicrosystems.neuralsignals",
+	"relatedMediaType": null,
+	"synonym": [
+    "Cerebus neural signal system",
+	"Blackrock Neural Signal 9"
```
lzehl: correct indent, change to "Blackrock Neural Signals 9"
I did not know that this existed... what is different here?

instances/data/contentTypes/blackrockmicrosystems.x-parallelrecordings.jsonld
```diff
@@ -0,0 +1,25 @@
...
+		".ns7",
+		".ns8",
+		".ns9",
+		".nev"
+	],
+	"name": "application/vnd.blackrockmicrosystems.parallelrecordings",
+	"relatedMediaType": null,
+	"synonym": [
+    "Cerebus neural signal system",
+	"Blackrock Parallel Recordings"
```
lzehl: correct indent

instances/data/contentTypes/x-ipynb+json.jsonld
```diff
@@ -0,0 +1,16 @@
+{
+	"@context" : {
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/x-ipynb+json",
```
lzehl, line 5: this is json based? I didn't know that :sweat_smile:

instances/data/contentTypes/x-matlab.jsonld
```diff
@@ -0,0 +1,15 @@
...
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/text/x-matlab",
+	"@type": "https://openminds.ebrains.eu/core/ContentType",
+	"fileExtension": [
+		".m"
+	],
+	"name": "text/x-matlab",
+	"relatedMediaType": null,
+	  "synonym": [
+    "MATLAB"
```
lzehl: correct indent

lzehl (changes requested):
only some small issues. @apdavison can you also check again?

instances/data/contentTypes/blackrockmicrosystems.neuralevents.jsonld
jagru20: additionally also change "Cerebus neural signal system" to  "Cerebus neural event system"?

lzehl: yes, I guess that makes sense. I wasn't sure where you had that synonym from... I thought that might have been how Cerebus refers to it...

instances/data/contentTypes/blackrockmicrosystems.neuralsignals.8.jsonld
jagru20: Same as for .ns9: in the specification ([link](https://blackrockneurotech.com/research/wp-content/ifu/LB-0023-7.00_NEV_File_Format.pdf)) it says on page 15: 

> In addition to the .NEV file described above, there can also be continuously sampled data files. These files have the extension .NSx where x is some number between 1 and 9. 

That's why they are in here.

instances/data/contentTypes/blackrockmicrosystems.neuralsignals.9.jsonld
jagru20: Same as for .ns8: in the specification ([link](https://blackrockneurotech.com/research/wp-content/ifu/LB-0023-7.00_NEV_File_Format.pdf)) it says on page 15: 

> In addition to the .NEV file described above, there can also be continuously sampled data files. These files have the extension .NSx where x is some number between 1 and 9. 

That's why they are in here.

instances/data/contentTypes/blackrockmicrosystems.neuralsignals.7.jsonld
jagru20: In the specification (link) it says on page 15:

>In addition to the .NEV file described above, there can also be continuously sampled data files. These files have the extension .NSx where x is some number between 1 and 9.

instances/data/contentTypes/x-ipynb+json.jsonld
jagru20: 😄 It actually is as the specification says: https://nbformat.readthedocs.io/en/latest/format_description.html 

I'll add that link to the jsonld and correct the indent.

instances/data/contentTypes/blackrockmicrosystems.neuralevents.jsonld
jagru20: I'll check that

jagru20: As far as I can tell, this term refers to hardware rather than software, should we then delete it and only keep the blackrock-term? I think that would be clear enough. And I think .nsx and .nev should not share a synonym just because they are recorded by the same hardware.

lzehl: @jagru20 yes then delete it. We'll pick it up again once we gather hardware metadata ;)

lzehl: you mean deleting it for all blackrock formats correct?

jagru20: yes, that's what I mean.

lzehl: :+1:

lzehl (approved):
Looks good to me. @jagru20 thanks for all this work you had with this large PR. 
@apdavison you think we can merge?

apdavison (approved):

apdavison:
Excellent work @jagru20 - ok to merge.

jagru20:
thank you two for all the feedback and patience! And for the kind words.

PR #217, opened 2021-08-04
Intro controlled terms + term suggestion for keywords

lzehl:
Intro controlled terms + term suggestion for keywords to further increase data integration
adding other taxons (postponed to later point); but in controlledTerms BiologicalOrder etc, will be added as study targets


lzehl:
@olinux remaining update for changing the property name in FileBundle (groupedBy -> groupingType) is already done in v4 without the PR

lzehl:
@olinux could you review and merge if everything is fine?

UlrikeS91 (approved):

PR #221, opened 2021-08-13
Update behavioralTask.schema.tpl.json

lzehl:
short update to rather include internalIdentifier as it is done for specimen and annotations etc. 
@olinux this will require an KGE update...

### DIFF
--- a/schemas/research/behavioralTask.schema.tpl.json
+++ b/schemas/research/behavioralTask.schema.tpl.json
@@ -14,7 +14,7 @@
     },
     "name": {
       "type": "string",
-      "_instruction": "Enter a descriptive full name for this behavioral task."
+      "_instruction": "Enter a descriptive name for this behavioral task."
     },
     "internalIdentifier": {
       "type": "string",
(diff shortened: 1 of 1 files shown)


schemas/research/behavioralTask.schema.tpl.json
```diff
@@ -5,20 +5,20 @@
...
   ],
   "properties": {
     "description": {
       "type": "string",
       "_instruction": "Enter a description of this behavioral task."
     },
-    "fullName": {
+    "name": {
       "type": "string",
       "_instruction": "Enter a descriptive full name for this behavioral task."
```
UlrikeS91: "Enter a descriptive name for this behavioral task." (without "full")

UlrikeS91 (changes requested):

schemas/research/behavioralTask.schema.tpl.json
lzehl: corrected

UlrikeS91 (approved):

PR #223, opened 2021-08-24
Research updates (behavioralTask/protocol/protocolExecution)

UlrikeS91:
Following a discussion with @lzehl  and @Peyman-N, we concluded on some updates for some of the `core/research` schemas:

 1. `behavioralTask` has been renamed to `behavioralProtocol`
 2. `studyOption` was removed from `protocol` since the `studyTarget` of the `protcolExecution` should cover this sufficiently
 3. A new schema, called `stimulation` has been added, which also has been added as property to both `behavioralProtocol` and `protocol` (in both cases with count 0 - N). This new schema has the following properties:
```
stimulationApproach | 1-1 | controlledTerms/stimulationApproach
stimulusType | 0 - 1 | controlledTerms/stimulusType  
```
`stimulationApproach` is similar to `experimentalApproach` giving an overarching type of stimulation/stimulus (e.g. auditory stimulus, electrical stimulus, etc.). @Peyman-N has added this as `stimulusModality` to the `controlledTerms` schemas. @Peyman-N, @lzehl and @apdavison, if you think that `stimulusModality` is more suitable than `stimulationApproach`, please let me know and I'll change it. 
`stimulusType` is similar to `technique` giving a more specific type of stimulus (e.g. direct current, static grating, etc.).  

@olinux FYI


schemas/research/stimulation.schema.tpl.json
```diff
@@ -0,0 +1,20 @@
+{
+  "_type": "https://openminds.ebrains.eu/core/Stimulation",
+  "required": [
+    "stimulationApproach"
```
lzehl: let's make type also a required one

lzehl (changes requested):

schemas/research/stimulation.schema.tpl.json
UlrikeS91: done

lzehl (approved):

PR #224, opened 2021-08-27
intro separate units for min / max value

lzehl:
in particular for age measurements this will be more convenient (issue was discussed with Oli)  
Note: we decided against nesting with QuantitativeValue (as we originally implemented the range) to avoid the nesting here, although from the graph perspective this would be cleaner.

### DIFF
--- a/schemas/miscellaneous/quantitativeValueRange.schema.tpl.json
+++ b/schemas/miscellaneous/quantitativeValueRange.schema.tpl.json
@@ -9,12 +9,18 @@
       "type": "number",
       "_instruction": "Add the maximum value measured for this range."
     },
+    "maxValueUnit": {
+      "_instruction": "Add the unit of measurement for the maximum value of this quantitative value range.",
+      "_linkedTypes": [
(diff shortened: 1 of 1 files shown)


schemas/miscellaneous/quantitativeValueRange.schema.tpl.json
```diff
@@ -9,12 +9,18 @@
...
+      ]
+    },
     "minValue": {
       "type": "number",
       "_instruction": "Add the minimum value measured for this range."
     },
-    "unit": {
-      "_instruction": "Add the unit of measurement of this quantitative value range.",
+    "minValueUnit": {
+      "_instruction": "Add the unit of measurement for the maximum value of this quantitative value range.",
```
UlrikeS91: needs to be minimum value in the instructions

UlrikeS91 (changes requested):

schemas/miscellaneous/quantitativeValueRange.schema.tpl.json
lzehl: done

UlrikeS91 (approved):

PR #227, opened 2021-09-06
Update quantitativeValueRange v3

UlrikeS91:
Recover old version of schema because PR #224 was merged into wrong version branch.


lzehl:
thanks for fixing this!

PR #231, opened 2021-09-13
Change requires --> required in a few places

apdavison:
I presume this should also be applied to v4


lzehl (approved):

lzehl:
@apdavison we need to correct this also in v4 (I just realized...)

PR #238, opened 2021-09-20
Removing component again

UlrikeS91:
FYI @olinux


lzehl (approved):
@UlrikeS91 thanks for taking care of this.

For everyone: sorry for the back and forth today... we'll find a better solution for registration / identification of research product collections in the near future.

PR #243, opened 2021-10-19
DSV "studiedSpecimen" change to "studiedSpecimenState"

UlrikeS91:
For less ambiguous linkage for specimen since a specimen could have been studied in 2 different dataset versions but in different states. With the specimen itself attached to the dataset version, it is not clear which state was the relevant one/the studied one.


UlrikeS91:
This has been solved differently now. Therefore, I close this PR.

PR #263, opened 2021-11-19
fixing contentType names for Blackrock formats

lzehl:
It seems that all the Blackrock NSX formats have the wrong name (they miss the number).
This PR fixes this.


jagru20 (approved):
True, this must be fixed. I do not see any further issues.

PR #264, opened 2021-11-19
Create yaml.jsonld

lzehl:


apdavison (commented):
I guess you found the same discussions I did: 
- https://stackoverflow.com/questions/332129/yaml-media-type
- https://www.quora.com/What-is-the-correct-MIME-type-for-YAML-documents

I think I would go for "application/yaml", for symmetry with "application/json", and because YAML is likely to ultimately get an official media type, but "application/vnd.yaml" is also fine.

lzehl:
Hi @apdavison yes I saw those and then this:
https://mailarchive.ietf.org/arch/msg/media-types/DA8UuKX2dyaVxWh-oevy-t3Vg9Q/

But I think you are right. For symmetry let's do "application/yaml"

lzehl:
Done :slightly_smiling_face: are the synonyms then correct ?

apdavison (approved):

PR #266, opened 2021-11-20
adding contentType for JSONLD

lzehl:
please double check the IANA reference

### DIFF
--- a/instances/data/contentTypes/ld+json.jsonld
+++ b/instances/data/contentTypes/ld+json.jsonld
@@ -1,0 +1,15 @@
+
+
+{
+	"@context" : {
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/ld+json",
+	"@type": "https://openminds.ebrains.eu/core/ContentType",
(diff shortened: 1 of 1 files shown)


lzehl:
@apdavison is this ready to be merged?

instances/data/contentTypes/ld+json.jsonld
```diff
@@ -0,0 +1,15 @@
...
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/ld+json",
+	"@type": "https://openminds.ebrains.eu/core/ContentType",
+	"fileExtension": [
+		".jsonld"
+	],
+	"name": "application/ld+json",
+	"relatedMediaType": "https://www.iana.org/assignments/media-types/application/ld+json",
+	"synonym": [
+	    "JSONLD"
```
apdavison: I think the synonym should be "JSON-LD", it's written that way almost everywhere.

apdavison (changes requested):

instances/data/contentTypes/ld+json.jsonld
lzehl: done :)

apdavison (approved):

PR #269, opened 2021-11-28
Create odml.jsonld

lzehl:


lzehl:
@apdavison OR @jagru20 I think it is enough if one of you reviews this PR. whoever is faster ;)

instances/data/contentTypes/odml.jsonld
```diff
@@ -0,0 +1,17 @@
+{
+	"@context" : {
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
+	"@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.odml",
```
jagru20: otherwise, if the specification is introduced/maintained by G-Node as the "specification" property's value suggests, I would use ``application/vnd.g-node.odml``

instances/data/contentTypes/odml.jsonld
```diff
@@ -0,0 +1,17 @@
...
+	"@type": "https://openminds.ebrains.eu/core/ContentType",
+	"fileExtension": [
+		".odml"
+	],
+	"name": "application/vnd.odml",
```
jagru20: see above

jagru20 (changes requested):
I suggest naming it `application/vnd.g-node.odml` because it seems to be introduced and maintained by G-Node, right?

instances/data/contentTypes/odml.jsonld
lzehl: yes. good point. done

instances/data/contentTypes/odml.jsonld
lzehl: done

jagru20 (approved):

lzehl:
@jagru20 feel free to merge (@apdavison that's okay, no?)

jagru20:
one more thing that I just noticed: shouldn't this PR go against the v4 branch?

PR #272, opened 2021-11-30
Moving recording to core

Peyman-N:
Moving the recording from the openMINDS-ephys to openMINDS-core.


UlrikeS91 (changes requested):
@lzehl and I worked on this together again

The concept of the schemas needs to be revised. It looks to us that the purpose of these schemas are to describe additional information about the actual data files from an (ephys) recording. If this is true, it would be quite comparable to data files that are images. They do have additional information that we want to capture in the future. 

In the past, the indirect linkage from an activity to a file via such a schema was discussed and rejected. The reason is the following: For high-level descriptions, this information would not be provided and a direct linkage from an activity to a file is needed. To allow for the extension with in-depth information about the data files, we decided to attach this additional information to those files (basically file "attributes"). This approach is not yet implemented, though. 

Therefore, we suggest to delete this two schemas and use some of the properties for attribute schemas of files (in this case for ephys recording files). Note: devices, software, etc. should be linked via an activity/content type and not directly to a file.

apdavison:
Hi @UlrikeS91,

The problem with your proposal is where an activity involves multiple devices, each recording data to a different file.

Here we need some way to connect a specific device to a specific file: attaching everything only to the activity does not provide this information.

To avoid indirect linking of the File to the Activity, I propose that we link both the File and the Recording to the activity, and then link the Recording to the File. This still supports the distinction between high-level and in-depth representations,  and has the additional advantage of being more consistent with the w3c Prov ontology, and the concept of qualifiedGeneration.

lzehl:
@apdavison could you provide an example for an activity that involves different devices that each produce a different file?

Independent of that: a triple linkage between activity - file - recording would work for me as well. This still means that the schemas need to be revised and only contain attributes of the file. And that would also mean that one recording can only be produced by one device or setup and one software.

apdavison:
@lzehl dual patch clamp, simultaneous calcium imaging and electrophysiology recording....  In some cases modern recording systems would store all data in a single file, but older setups are likely to use independent systems for the different devices.

lzehl:
@apdavison I would register all of the mentioned examples in different activities (each with a single responsible device/software) and define their simultaneous nature between them via an in-depth schema. If all activities with different devices store data within one file it would indeed require to link the triplet activity (a) -> file -> recording <- activity (a), but the device would remain linked on the activity not the file or the recording OR we could link activity (a) -> file -> [recording | image] and activity (a) -> device 'state' (r) -> recording | activity (a) -> device 'state' (i) -> image. What do you think?

schemas/data/channel.schema.tpl.json
```diff
@@ -0,0 +1,16 @@
...
+            "_instruction": "Enter any internal identifier used for this channel."
+        },
+        "unitOfMeasurement": {
+            "_instruction": "Add the unit of measurement for this channel.",
+            "_linkedTypes": [
+                "https://openminds.ebrains.eu/controlledTerms/UnitOfMeasurement"
+            ]
+        }
+    }
+}
```
lzehl: newline missing at the end of the JSON

schemas/data/recording.schema.tpl.json
```diff
@@ -0,0 +1,67 @@
+{
+    "_type": "https://openminds.ebrains.eu/core/Recording",
+    "required": [
+        "timeStep"
```
lzehl: is this really the only required field we can ask for? I think there are properties missing here (e.g. channels?)

schemas/data/channel.schema.tpl.json
```diff
@@ -0,0 +1,16 @@
+{
+    "_type": "https://openminds.ebrains.eu/core/Channel",
+    "required": [],
```
lzehl: if there is nothing required this schema is not really helpful I think. unitOfMeasurement has to be required

schemas/data/recording.schema.tpl.json
```diff
@@ -0,0 +1,67 @@
...
+                "https://openminds.ebrains.eu/core/Channel"
+            ]
+        },
+        "timeStep": {
+            "_instruction": "Time step (inverted sampling frequency) of data",
+            "_embeddedTypes": [
+                "https://openminds.ebrains.eu/core/QuantitativeValue"
+            ]
+        },
+        "order": {
```
lzehl, line 26: the use case is unclear to me for this property. could you explain it in more detail?

schemas/data/recording.schema.tpl.json
```diff
@@ -0,0 +1,67 @@
...
+            "minItems": 1,
+            "uniqueItems": true,
+            "_instruction": "Add the locations of the files / file bundles in which the recorded data are stored",
+            "_embeddedTypes": [
+                "https://openminds.ebrains.eu/core/File",
+                "https://openminds.ebrains.eu/core/FileBundle"
+            ]
+        }
+    }
+}
```
lzehl: newline is missing at the end of the JSON

schemas/data/recording.schema.tpl.json
```diff
@@ -0,0 +1,67 @@
...
+            "uniqueItems": true,
+            "_instruction": "Add all the devices used for this recording",
+            "_linkedTypes": [
+                "https://openminds.ebrains.eu/core/hardwareVersion",
+                "https://openminds.ebrains.eu/ephys/Pipette",
+                "https://openminds.ebrains.eu/ephys/ElectrodeArray",
+                "https://openminds.ebrains.eu/ephys/Electrode"
+            ]
+        },
+        "unitOfMeasurement": {
```
lzehl: this can't be stated here and in channel. this should only be asked for in channel

schemas/data/recording.schema.tpl.json
```diff
@@ -0,0 +1,67 @@
...
+    "_type": "https://openminds.ebrains.eu/core/Recording",
+    "required": [
+        "timeStep"
+    ],
+    "properties": {
+        "internalIdentifier": {
+            "type": "string",
+            "_instruction": "Enter any internal identifier used for this recording"
+        },
+        "channels": {
```
lzehl: should be "channel" not "channels"

schemas/data/recording.schema.tpl.json
```diff
@@ -0,0 +1,67 @@
...
+            "items": {
+                "type": "integer"
+            },
+            "_instruction": "If this recording is part of a sequence (sweeps, repetitions), enter the order(s) in that sequence contained in this recording"
+        },
+        "description": {
+            "type": "string",
+            "_instruction": "Enter a description of this recording."
+        },
+        "devices": {
```
lzehl, line 39: this needs to be linked from channel not from here I think.

schemas/data/recording.schema.tpl.json
```diff
@@ -0,0 +1,67 @@
...
+                "https://openminds.ebrains.eu/ephys/Electrode"
+            ]
+        },
+        "unitOfMeasurement": {
+            "_instruction": "Add the unit of measurement of this recording, if the unit applies to all channels. If different channels have different units, provide them in the `channels` field.",
+            "_linkedTypes": [
+                "https://openminds.ebrains.eu/controlledTerms/UnitOfMeasurement"
+            ]
+        },
+        "dataLocation": {
```
lzehl: should be "definedIn" or "storedIn" to be more consistent with our other schemas

lzehl (changes requested):
@Peyman-N thank you for the revision but this does not yet match what we discussed offline from the graph model if I'm not mistaken.

Fact sheet we put together in this discussion:
+ A recording is a more detailed description of the content of a file.
+ A file can store one or multiple recordings.
+ A recording can contain one or multiple channel. 
+ A recording could be distributed over multiple files.
+ Each channel can originate from a device "state" (of one contact point of a single electrode, of a pipette, of one contact point of an electrode of an electrode array).
+ "Recording" should probably be renamed to "EphysRecording"

Resulting graph model we sketched:
+ SpecificActivity (e.g. patch clamping) links to Device"States" and to Files
+ Device"State" link to Channel of Recording
+ Recording also link to Files
+ graph validation point: Files of Recording/ Files of SpecificActivity

Does this summarize our discussion?

schemas/data/channel.schema.tpl.json
Peyman-N: I agree about unit Of Measurement have to be required but I think internal Identifier should be required too. Having a channel without any identification is ambiguous.

lzehl: I agree

schemas/data/recording.schema.tpl.json
Peyman-N: I think "definedIn" should be requied as having a recording without a file attached to it is seems incomplete. @apdavison what are your thoughts on this?

Peyman-N:
This pull request is obsolete as in an offline meeting between @apdavison, @UlrikeS91, @lzehl and @Peyman-N; it was decided to move the recording back to ephys. 
Any the new contribution should be on pull request [#31](https://github.com/HumanBrainProject/openMINDS_ephys/pull/31) of ephys

PR #277, opened 2021-12-16
adding EBRAINS terms of use

lzehl:


lzehl:
@olinux FYI as discussed.

UlrikeS91 (approved):

PR #280, opened 2022-01-03
first attempt to formalize graph dependencies

olinux:
Hello,
As discussed previously, we need a way on how to formalize the dependencies of various properties in the graph so tools (such as the KG Editor and validation mechanisms) can make use of it. This is a first attempt to do so -> the syntax is kept close to JSON-LD as well as the EBRAINS KG query language

### DIFF
--- a/schemas/data/file.schema.tpl.json
+++ b/schemas/data/file.schema.tpl.json
@@ -53,6 +53,19 @@
       "_instruction": "Add one or several bundles in which this single file can be grouped.",
       "_linkedTypes": [
         "https://openminds.ebrains.eu/core/FileBundle"
+      ],
+      "_restrictions": [
+        {
+          "_isPartOf": [
+            {
+              "@id": "fileRepository"
+            },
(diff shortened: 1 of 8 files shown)


lzehl:
@olinux thanks for starting to tackle this. Can you explain the syntax you used a bit more though? I have the feeling we should simplify it a bit more for the openMINDS syntax (if possible). Let me know if you want to set up a meeting to discuss this (might be easier than explaining/discussing it here).

olinux:
I do agree that this is slightly more complex than the definition of properties, but I also assume, that this is something which isn't handled by the majority of contributors but rather a few "experts". I think it's easier to go through this in a meeting to explain the thoughts behind it.

lzehl:
This PR should be further discussed in an issue on main https://github.com/openMetadataInitiative/openMINDS/issues/68. @Raphael-Gazzotti @tgbugs @MBAbrams started re-discussing this a bit.

PR #298, opened 2022-02-03
logic reassessment of file bundle groupings

lzehl:
The content of a file bundle is defined by a file path pattern.

The file path pattern defines the overall grouping mechanism (grouping type).
@olinux : there is a graph dependency here, because the grouping type of the file path pattern has to match the one of the file bundle.

Assuming that "study target" was selected as grouping type.
The old "descendedFrom" property does not allow any further specifications regarding study targets.

To include them, I changed the property name to "groupedBy" (to better fit sementically) and added the studyTarget as linked category. 

In addition, I changed the "groupedBy" property back to be an array. Based on how the file bundles are generated, they could be based on multiple entities of the same grouping type. This is particularly true for study targets (e.g., study target grouping could be based on specific area + cell type).
@olinux this includes again a graph validation: in "groupedBy" only entities are allowed of the specified grouping type.

To go even further (and becoming more flexible): the grouping type could be an array as well, allowing to create file bundles that could be grouped by e.g. a subject and a study target. 

@HumanBrainProject/openminds-developers what do you think? I will adopt this PR draft based on our discussion.

### DIFF
--- a/schemas/data/fileBundle.schema.tpl.json
+++ b/schemas/data/fileBundle.schema.tpl.json
@@ -16,10 +16,11 @@
       "type": "array",
       "minItems": 1,
       "uniqueItems": true,
-      "_instruction": "Add all entities that played a role in the production of this file bundle.",
+      "_instruction": "Add all entities that defined which file was grouped into this file bundle.",
       "_linkedCategories": [
         "fileOrigin",
-        "studyTarget"
+        "studyTarget",
+        "coordinateSpace"
       ]
     },
     "format": {
--- a/schemas/data/fileBundle.schema.tpl.json
+++ b/schemas/data/fileBundle.schema.tpl.json
@@ -12,10 +12,14 @@
       "type": "string",
       "_instruction": "Enter a short content description for this file bundle."
     },
-    "descendedFrom": {
-      "_instruction": "Add the entity that played a role in the production of this file bundle.",
+    "groupedBy": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add all entities that played a role in the production of this file bundle.",
       "_linkedCategories": [
-        "fileOrigin"
+        "fileOrigin",
+        "studyTarget"
       ]
     },
     "format": {
--- a/schemas/data/fileBundle.schema.tpl.json
+++ b/schemas/data/fileBundle.schema.tpl.json
@@ -33,7 +33,7 @@
       "type": "array",
       "minItems": 1,
       "uniqueItems": true,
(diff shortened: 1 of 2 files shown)


UlrikeS91:
I think the draft looks good and the general idea behind it. It will increase the complexity of potential groupings, but that might not be a bad thing. 

However, I don't understand what you meant by this:
> To go even further (and becoming more flexible): the grouping type could be an array as well, allowing to create file bundles that could be grouped by e.g. a subject and a study target.

Could you elaborate on this? Do you mean the property "groupingType" which links to the controlledTerm/fileBundingGrouping? If so, how would this look like?

lzehl:
@UlrikeS91 yes I should give a better example here. If the groupingType for file path patterns and therefore also file bundles would be an array you could do the following (example):

+ groupingType: "subject", "study target"
-> FileBundle A groupedBy: subject A, study target X
-> FileBundle B groupedBy: subject A, study target Y
-> FileBundle C groupedBy: subject B, study target X
-> FileBundle C groupedBy: subject B, study target Y

@olinux the regex (file path pattern) logic already allows for multiple groupings correct?

UlrikeS91:
Thanks for the additional explanation. This seems reasonable to me. What does the rest of the @HumanBrainProject/openminds-developers think?

lzehl:
@HumanBrainProject/openminds-developers it would be good to hear more opinions on the matter. I would like to implement this change as soon as possible if we agree on a proper approach.

Here also another small extension we should consider for this change:

We should rename the category "coordinateSpace" to groupedBy and add this category to CommonCoordinateSpace and CustomCoordinateSpace in SANDS. And respectively adding "coordinate space" to FileBundleGrouping.

lzehl:
@UlrikeS91 & @olinux please review the PR carefully and comment on/approve the following action points:

**Changes in filePathPattern.schema.tpl.json:**
+ Property "groupingType" is now an array.
+ Instructions of property "regex" were changed/extended.
**---> This should not have an effect on the current entries in the EBRAINS KG, correct?**

**Changes in fileBundle.schema.tpl.json**
+ Property "descendedFrom" is renamed into "groupedBy" and changed into an array allowing linkages of schemas that belong to the categories "fileOrigin", "studyTarget", or "coordinateSpace". Instructions were also refined.
+ Property "groupingType" is now an array. Instructions were refined accordingly.
**---> This will effect on the current entries in the EBRAINS KG!**
**---> Action: transferring all entries of deprecated property "descendedFrom" to "groupedBy".**

lzehl:
@UlrikeS91 please double check but do not merge yet even if it's okay (just state your okay). I want to make sure with @olinux how this is affecting EBRAINS (and what needs to be done there to integrate this change)

UlrikeS91 (approved):

PR #300, opened 2022-02-03
adding new contentTypes related to TVB pipeline

lzehl:
relevant for https://search.kg.ebrains.eu/instances/4edef2ff-e653-40a5-a222-0a5759fdaaaf
+ json (specific for old MINDS, as far as formerly specified)
+ tsv (specific for TVB)

### DIFF
--- a/instances/data/contentTypes/minds+json.jsonld
+++ b/instances/data/contentTypes/minds+json.jsonld
@@ -1,0 +1,14 @@
+{
+	"@context" : {
+		"@vocab": "https://openminds.ebrains.eu/vocab/"
+	},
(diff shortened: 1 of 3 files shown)


lzehl:
@jagru20 / @apdavison : I do not know who of you has more time at the moment ;) (one of you for the review is enough I guess)

apdavison (commented):
I think it would be better to spell out tvb in full, and also to specify what kind of data are being stored, because I presume TVB works with multiple different formats, i.e. `application/vnd.thevirtualbrain.connectivity+tsv` (where "connectivity" would be replaced with whatever is actually in the files)

lzehl:
@apdavison yes you are right. I'll change it

lzehl:
done :)

apdavison (approved):

Issue #292, opened 2022-02-02
adding QuickNII contentTypes


lzehl:
**ContentType schema:** 
+ https://humanbrainproject.github.io/openMINDS/v3/ **--> select** core/DATA/ContentType
+ https://github.com/HumanBrainProject/openMINDS_core/blob/v3/schemas/data/contentType.schema.tpl.json

**required property:** name
**recommended property:** extension, as well as description AND/OR link to specification

**Missing contentTypes (suggested names):**
+ application/vnd.quicknii+json
+ application/vnd.quicknii+xml
+ application/vnd.quicknii+flat

UlrikeS91:
@Tevemadar and @majpuc can be asked for further questions

PR #302, opened 2022-02-04
Add QuickNII contentTypes

MaaikevS:
Related to Issue #292.

Please let me know if anything needs to be changed


instances/data/contentTypes/quicknii+flat.jsonld
```diff
@@ -0,0 +1,15 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.quicknii+flat",
```
lzehl: @jagru20 should it be `vnd.quicknii+flat` OR `vnd.quicknii.flat` ? I wonder since `flat` is not a general format with an official media type...

lzehl (commented):
@MaaikevS thanks a lot for this PR. I leave my comment to be answered by @jagru20

instances/data/contentTypes/quicknii+flat.jsonld
jagru20: In that case I would go for `vnd.quicknii.flat` if I remember correctly, the + suffix should only be used with well-known container formats (e.g., xml, zip) that are also well known beyond the neurosience community.

MaaikevS: Thank you @lzehl @jagru20. I will change the file name to `vnd.quicknii.flat`

lzehl (approved):

lzehl:
@jagru20 thanks for the quick feedback :+1: 
@MaaikevS thanks for pushing this PR :slightly_smiling_face: 

I'm merging now :wink:

lzehl:
@MaaikevS FYI I will do a rebase to v4 (the latest version of openMINDS; this error happend a lot to me and others as well... we need to switch the default branch soon).

MaaikevS:
@lzehl Thanks! I hadn't noticed it was on v3 and not v4. Will make sure to check next time ;)

PR #303, opened 2022-02-08
update short name instruction

lzehl:
... to loosen the instructions (in relation to format) and define better its purpose (for what it should be used)


UlrikeS91 (approved):

UlrikeS91:
Looks good and makes sense. I did the same update on the only other schema that also uses shortName - commonCoordinateSpace in SANDS. See the mentioned PR in the thread here (or @lzehl you are also set as reviewer).

PR #304, opened 2022-02-08
Update BIDS content Types

lzehl:
TODOs in the PR:
+ adding IRI to BIDS spec
+ updating names
+ adding missing content types

QUESTION: how detailed do we want to be here?
A) Each TSV and JSON type becomes its own content type?
B) Should we have one content type for all TSV and one content type for all JSONs?

@apdavison @Peyman-N and @jagru20 please advise before I continue


UlrikeS91:
I'm closing this PR. I made an issue about this in the instances-repo and will create a new PR when a conclusion has been reached.

PR #305, opened 2022-02-09
Update contentType.schema.tpl.json with displayLabel

lzehl:
The names of the contentTypes are highly accurate but also highly technical and sometimes quite long. 
For better display in service systems such as the KG Search I suggest to add a (optional) displayLabel property.

### DIFF
--- a/schemas/data/contentType.schema.tpl.json
+++ b/schemas/data/contentType.schema.tpl.json
@@ -4,6 +4,10 @@
     "name"  
   ],
   "properties": {
+    "displayLabel": {
+      "type": "string",
+      "_instruction": "Enter a display label for this content type."
+    },
     "fileExtension": {
       "type": "array",
       "minItems": 1,


UlrikeS91 (approved):

apdavison:
As an alternative, how about taking the first entry in the "synonym" list?

lzehl:
@apdavison question is if we can rely on this to look nice. but generally speaking that was my other suggestion. the display label property is more explicit though (in relation to the purpose of the additional name).

apdavison:
@lzehl we would have to check that all entries have suitable entries in `synonym[0]`, but that's slightly less work than adding `displayLabel` to each entry.

I'm arguing against `displayLabel` (although not very strongly) because I don't think the openMINDS structure should be influenced by UI considerations.

lzehl:
@apdavison I see your point. maybe we can change the naming to "preferredLabel" to make it a bit independent of UI. The synonym[0] option is too implicit I would argue.

lzehl:
@UlrikeS91 could you take over and change the prop name?

UlrikeS91:
> @UlrikeS91 could you take over and change the prop name?

Yes, can do that.

PR #310, opened 2022-02-25
Update strain.schema.tpl.json

UlrikeS91:
Just noticed that "species" in the strain schema was optional. Checked the closed issues and such, but there we discussed that this needs to be required because specimen can either choose CT/species or core/strain as "species", but the species information must be there (either directly or via strain schema and then species).


UlrikeS91:
@lzehl this is still open - Could you review the PR?

lzehl (approved):

PR #311, opened 2022-02-25
Added VisuAlign contentType

MaaikevS:


instances/data/contentTypes/visualign+json.jsonld
```diff
@@ -0,0 +1,16 @@
...
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.visualign+json",
+  "@type": "https://openminds.ebrains.eu/core/ContentType",
+  "fileExtension": [
+    ".json"
+  ],
+  "name": "application/vnd.visualign+json",
+  "relatedMediaType": "https://www.iana.org/assignments/media-types/application/json",
```
jagru20, line 11: I am not sure what exactly qualifies as a related mediatype, in the past, we only linked an IANA type here, if it matched exactly, because it says that in the documentation, but I would be fine with linking a "best-fit" Media type. @apdavison what do you think?

jagru20:
based on [this comment ](https://github.com/HumanBrainProject/openMINDS_core/pull/320#issuecomment-1073669832) I would take this as-is.

PR #309, opened 2022-02-25
Add ilastik content types

MaaikevS:
@lzehl @UlrikeS91 

I added the content types for Ilastik based on Issue #296 
If you find that anything is missing or incorrect, let me know.

PR #320, opened 2022-03-17
Create vnd.espina.seg+ms-excel and ...+txt

lzehl:
homepage is https://cajalbbp.es/espina/ (but unfortunately I did not find any real specification for the formats)

### DIFF
--- a/instances/data/contentTypes/vnd.espina.seg+ms-excel.jsonld
+++ b/instances/data/contentTypes/vnd.espina.seg+ms-excel.jsonld
@@ -1,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.espina.seg+ms-excel",
+  "@type": "https://openminds.ebrains.eu/core/ContentType",
+  "fileExtension": [
+    ".xls"
+  ],
+  "name": "application/vnd.espina.seg+ms-excel",
+  "relatedMediaType": null,
+  "specification": null,
+  "synonym": null
+}
--- a/instances/data/contentTypes/vnd.espina.seg+txt.jsonld
+++ b/instances/data/contentTypes/vnd.espina.seg+txt.jsonld
@@ -1,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.espina.seg+txt",
+  "@type": "https://openminds.ebrains.eu/core/ContentType",
+  "fileExtension": [
+    ".txt"
+  ],
+  "name": "application/vnd.espina.seg+txt",
+  "relatedMediaType": null,
+  "specification": null,
+  "synonym": null
+}


instances/data/contentTypes/vnd.espina.seg+ms-excel.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.espina.seg+ms-excel",
```
jagru20, line 5: I also commented this in a different PR: I do not think that single-file formats as excel, txt, csv, etc. qualify as container formats. According to the naming convention as mentioned [here ](https://wiki.ebrains.eu/bin/view/Collabs/content-types/Naming%20convention/?srid=XHN9Kva0) the suffix is meant for container format. I would go for `application/vnd.espina.seg.ms-excel` instead

instances/data/contentTypes/vnd.espina.seg+txt.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.espina.seg+txt",
```
jagru20, line 5: same as above, I would go for `application/vnd.espina.seg.txt`, declaring it as a sub-subtype

lzehl:
@jagru20 and @apdavison  can you explain more what the differences are?

I don't understand the convention based on the examples we already have in. 
From those I understood the convention the following way, but maybe I'm wrong:
+ `a.b` is be used when `a` has the format subtype `b` and `b` is not based on any other format
+ `a+b` is used when `a` has the format subtype `b` which is a standalone base format (less structured than a+b; same extensions as b)
+ `a.b+c` is used when `a` has the format subtype `b` which is based on a standalone base format `c` (different extensions as c)

Which also means I messed up a bit the instances above... If I follow this idea it should have been `vnd.espina+txt` `vnd.espina+ms-excel` and (from another PR where I did it correctly) `vnd.espina.seg+zip`.

@MaaikevS & @UlrikeS91 FYI (related question to your PRs #311 #309 )

MaaikevS:
@lzehl Thank you for pointing this out. This is also seems related to the already merged PR #302 
I will change all of these content types if the names used are indeed not in line with the naming convention!

jagru20:
Sorry if I am causing confusion here. The naming convention [(link)](https://wiki.ebrains.eu/bin/view/Collabs/content-types/Naming%20convention/?srid=XHN9Kva0) proposed by Andrew states that 
"Where the format uses a standardised container format, such as XML, JSON, HDF5, we add `+{ container_format }`, e.g.
` application/vnd.nwb.nwbn+hdf`"
Whereas another `.{sub_sub_type}` shows that this is another subtype of the same supertype. 
This means, a + Suffix does not sign a subtype, only a container format that applies to this contenttype. In your example above, @lzehl I agree that there should be three types for espina, but not named as `vnd.espina+txt` and `vnd.espina+ms-excel` but rather `vnd.espina.ms-excel` and `vnd.espina.ms-excel` (better `vnd.espina.xls`) as these seems to be subtypes of espina, but not container types.

apdavison:
What is the difference between the two formats? Are they interchangeable (the same data can be stored in either), or do they have different roles (contain different types of data)?

lzehl:
@apdavison as far as I understand, the data in these formats hold different data information. They exist in parallel in datasets that are using Espina and even if (xls and txt do not have the same structure, but I can double check) are differently structured

apdavison:
ok, then I would name them `application/vnd.espina.purpose1` and `application/vnd.espina.purpose2` where "purpose1" and "purpose2" are the kind/type of data that they contain _or_ what those data are used for. If there is value in also being able to open the file in Excel then you could add `+ms-excel` to the one that is an Excel file (almost anything can be opened in a text editor, so I'm not sure adding "+txt" is valuable).

lzehl:
@apdavison thanks a lot for the feedback. I will try later also to document this for contributions to content types (I will pin you and @jagru20 on this for corrections). 

@jagru20 I will revise today the PR and pin you when I'm done with the corrections.

UlrikeS91:
I'm closing this PR. It is replaced by the new PR in the instances-repo.

PR #321, opened 2022-03-21
Extend the Stimulation schema

apdavison:
to allow attaching data files containing the stimulus content (e.g. movies, audio files, injected current traces), and to allow parameterised stimuli.

### DIFF
--- a/schemas/research/stimulation.schema.tpl.json
+++ b/schemas/research/stimulation.schema.tpl.json
@@ -38,7 +38,7 @@
       "type": "array",
       "minItems": 1,
       "uniqueItems": true,
-      "_instruction": "Add all the locations that the data attached to the entity is published",
+      "_instruction": "Add all data files containing the stimulus content (e.g. movies, audio files, injected current traces).",
       "_embeddedTypes": [
         "https://openminds.ebrains.eu/core/File",
         "https://openminds.ebrains.eu/core/FileBundle"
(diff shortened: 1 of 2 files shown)


schemas/research/stimulation.schema.tpl.json
```diff
@@ -24,6 +24,25 @@
...
+      "_instruction": "Add all important parameters grouped in context-specific sets that were used in this stimulus",
+      "_embeddedTypes": [
+        "https://openminds.ebrains.eu/core/ParameterSet"
+      ]
+    },
+    "dataLocation": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add all the locations that the data attached to the entity is published",
```
lzehl: what is meant here? the stimulation data that is applied during an experiment or the stimulation data recorded during the experiment or the data of interest that are recorded during a stimulation? 

The latter should be covered elsewhere, but I assume you mean the first two (one or both). Could you specify the instructions a bit more?

lzehl (changes requested):
@apdavison looks fine to me except the instruction for the data location in stimulation. This may lead to erroneous entries if we are not precise about what data is meant here.

lzehl (approved):

lzehl:
@apdavison I will merge this despite the fact that we probably should remove this from core entirely and set up something more in-depth fitting. I need more time to rethink this though so we might use this version nonetheless.

Issue #293, opened 2022-02-02
add MeshView contentTypes


lzehl:
**ContentType schema:** 
+ https://humanbrainproject.github.io/openMINDS/v3/ **--> select** core/DATA/ContentType
+ https://github.com/HumanBrainProject/openMINDS_core/blob/v3/schemas/data/contentType.schema.tpl.json

**required property:** name
**recommended property:** extension, as well as description AND/OR link to specification

**Missing contentTypes (suggested names):**
+ application/vnd.meshview+json
+ ??? (some text file)

UlrikeS91:
@Tevemadar and @majpuc can be asked for further questions

Majpuc:
For the Nutil Quantifier output json: The coordinate files are identical across the board, from Nutil v0.33 to v0.8.0.

MaaikevS:
As discussed in the PR #332, the meshview contentType mentioned in this issue is rather a Nutil contentType.
@Majpuc Unless there are other contentTypes for Meshview that need to be mapped, we can close this issue.

Majpuc:
Did we want to create a content type for the manually input [x1, y1, z1, x2, y2, z2,..., xn, yn, zn] values or not? If not, we can close this issue.

MaaikevS:
@lzehl I think you may want to create another contentType for the meshview input that is saved as a csv file?
The coordinates would be stored as follows:
![image](https://user-images.githubusercontent.com/31585951/175242924-de02674e-e179-44c3-9131-3314c2f54e33.png)
You would just copy paste the coordinates (so not the headers) into the "cloud from coordinates" field.

lzehl:
I would define it yes. Probably as application/vnd.meshview.coord+csv

What I do not understand how the CSV you show above match with the input structure you stated before ([x1, y1, z1, x2, y2, z2,..., xn, yn, zn]). Maybe this is just a misunderstanding... but the above CSV would translate to a list of list I think  ([[x1, y1, z1], [x2, y2, z2],..., [xn, yn, zn]]). @apdavison could you have a look? I may just get it wrong...

MaaikevS:
So the input structure we discussed above ([x1, y1, z1, x2, y2, z2,..., xn, yn, zn]) is for how the coordinates are stored in the meshview compatible json file. This is different from the way the CSV is organised.

Tevemadar:
While the format itself is under change at the moment, it looks like this:
```
[
   {
[...]

Issue #294, opened 2022-02-02
add Nutil contentType


lzehl:
**ContentType schema:** 
+ https://humanbrainproject.github.io/openMINDS/v3/ **--> select** core/DATA/ContentType
+ https://github.com/HumanBrainProject/openMINDS_core/blob/v3/schemas/data/contentType.schema.tpl.json

**required property:** name
**recommended property:** extension, as well as description AND/OR link to specification

**Missing contentTypes (suggested names):**
+ application/vnd.nutil.quantifier (extension: .nut)
+ application/vnd.nutil.transform (extension: .nut)
+ application/vnd.nutil+csv
+ image/vnd.nutil+png

UlrikeS91:
@leuat, @sharoncy and @majpuc: Are the .nut file for Nutil all the same regardless of which function is used (e.g. Transform vs. Quantifier)?

sharoncy:
No they are different. The .nut file structure also varies between Nutil versions.

UlrikeS91:
> No they are different. The .nut file structure also varies between Nutil versions.

Thank you for the rapid response, @sharoncy. 

So, this means that we will need to add 4 contentTypes for the nut file extension - one per function:
- Transform
- Quantifier
- TiffCreator
- Resize

Correct?

sharoncy:
I discussed this with Ulrike in person: 4 contentTypes for the nut file extensions is a good idea. To fit with the OpenMINDS schema she suggested sharing the different Nutil features as different "software", each linked to a different contentType (so a different KG datacard for Nutil Quantifier, Transform, Resize and TiffCreator). This sounds like a good idea as the Nutil features are standalone. But this may have implications elsewhere, let's discuss @Majpuc and @leuat .

Majpuc:
Thank you for the info. This is going to be tricky as the software that are registrered in the KG  need to have corresponding descriptions in the TC documents. Currently, we have one ID for Nutil as a whole and one ID for Nutilweb as a whole, because the web version will have a different access point and don't allow all features. They also use different content types. What do you think @leuat? We discussed this a lot during the meeting yesterday. I propose to have a meeting about this in order to give you a bit more background information.

sharoncy:
Yes please, I am not fully informed. @leuat @Majpuc

lzehl:
@sharoncy you said the nut files are also different between nutil versions? meaning how many versions should we cover for the 4 nut files?

sharoncy:
[...]

PR #332, opened 2022-06-07
Create meshview+json.jsonld

MaaikevS:
Related to #293 
There might be more content types that need to be added, but I need this one for a dataset

### DIFF
--- a/instances/data/contentTypes/meshview+json.jsonld
+++ b/instances/data/contentTypes/meshview+json.jsonld
@@ -1,0 +1,16 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.meshview+json",
+  "@type": "https://openminds.ebrains.eu/core/ContentType",
+  "fileExtension": [
+    ".json"
+  ],
+  "name": "application/vnd.meshview+json",
+  "relatedMediaType": "https://www.iana.org/assignments/media-types/application/json",
+  "synonym": [
+    "Meshview JavaScript Object Notation",
+    "Meshview JSON"
+  ]
+}
--- a/instances/data/contentTypes/nutil.quantifier+json.jsonld
+++ b/instances/data/contentTypes/nutil.quantifier+json.jsonld
@@ -2,15 +2,15 @@
   "@context": {
     "@vocab": "https://openminds.ebrains.eu/vocab/"
   },
-  "@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.meshview+json",
+  "@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.nutil.quantifier+json",
   "@type": "https://openminds.ebrains.eu/core/ContentType",
   "fileExtension": [
     ".json"
   ],
-  "name": "application/vnd.meshview+json",
+  "name": "application/vnd.nutil.quantifier+json",
   "relatedMediaType": "https://www.iana.org/assignments/media-types/application/json",
   "synonym": [
-    "Meshview JavaScript Object Notation",
-    "Meshview JSON"
+    "Nutil Quantifier JavaScript Object Notation",
+    "Nutil Quantifier JSON"
   ]
 }


MaaikevS:
@lzehl Would you be able to review this?

instances/data/contentTypes/meshview+json.jsonld
```diff
@@ -0,0 +1,16 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/contentTypes/application/vnd.meshview+json",
```
lzehl: what is it used for? what does it contain? maybe we can make the name a bit more specific (e.g. vnd.meshview.config+json)

lzehl (changes requested):
if changes are actually needed depends on the answer to my question ( see comment).
Formally the jsonld looks fine.

instances/data/contentTypes/meshview+json.jsonld
lzehl: if possible provide a small description (property "description") or add a link to the specification of this format (property "specification"

MaaikevS: Thanks for looking at this. The json file is the file that is uploaded to the meshview viewer, so it is for the visualisation of point coordinates. So would you want it to become vnd.meshview.visualization+json? Or something like this?

lzehl: Mh if this is an input file then I think it should be a contentType of the software that produced it (except you have to write it manually or it is actually also produced by MeshView). So not MeshView but?

MaaikevS: Hmm I don't know how it was produced initially, I will check. It has the nested structure and properties that are compatible with meshview

Majpuc: Hi, MeshView (e.g. https://meshview.apps.hbp.eu/?atlas=ABA_Mouse_CCFv3_2017_25um) has two types of input. 1. a json file created by Nutil. 2. The user is providing a coordinate list of points x,y,z by pasting them directly in the window under "clouds from coordinates" I hope this helps :)

MaaikevS: Thanks @Majpuc. Can you please advise us then on what the correct content types should be regarding Issues #294 and #293 
It would be great to finish those content types that they can actually be used in the curation!

Majpuc: I have asked the developer. Will come back to you asap

lzehl: @Majpuc I'm missing the JSON from the list we discussed for Nutil (cf. #294); could you clarify/verify that we need a JSON for Nutil? What is the content / structure of this file?

The second input seems to be somewhat MeshView specific; but the list is of what format? TXT, CSV, ? and what is the order of the list ( is it: [[x1, y1, z1],[x2, y2, z2], ...] OR [[x1, x2, ...], [y1, y2, ...], [z1, z2, ...]])?

MaaikevS: The JSON coordinates are stored [[x1, y1, z1],[x2, y2, z2], ...]
The "clouds from coordinates" is CSV or TSV but copy-pasted into the box

Majpuc: All the content types were defined in our common spreadsheet https://docs.google.com/spreadsheets/d/1e3-D-uCsMeYCSkjBmPC--jpOH1iJZWqcj58Mb6GB0T4/edit#gid=0
the json is a Nutil Quantifier output file.
The cloud from coordinate are listed [x1, y1, z1],[x2, y2, z2], ...

Majpuc: Concerning NutilQuantifier output jsons:  the coordinate files are identical across the board, from Nutil v0.33 to v0.8.0

MaaikevS: Actually, the coordinates are listed as one long list of x,y,z coordinates that are not stored as individual lists. My earlier response was incorrect. They are stored as [x1, y1, z1, x2, y2, z2,..., xn, yn, zn]. So they are extracted by meshview as triplets.

lzehl: this turns out to be very complicated :sweat_smile: 

I've got a bit lost now. Let's summarize:

MeshView has two different input Formats (A, B)
A: .json extension; should be the ContentType of Nutil Quantifier JSON
B: .tsv or .csv (as manual copy/paste); required structure  [x1, y1, z1, x2, y2, z2,..., xn, yn, zn] ??? (which would only be one line and no tab separation?); Could be considered as a MeshView specific input file.

MaaikevS: Sorry to make it more difficult. Let me explain what I know about the json files.
The coordinates in the json file are stored as follows: there is a list of values for "triplets". The values represent the x,y,z coordinates and are ordered like this:[x1, y1, z1, x2, y2, z2,..., xn, yn, zn] 
image

If you want to use the other option and copy paste the coordinates directly into meshview you just copy a list of values that are comma or tab separated, i.e. x1, y1, z1, x2, y2, z2,..., xn, yn, zn

Hope this helps.

lzehl: okay. I slowly start to get a picture. I would say the following:

The JSON has to be captured as Nutil Quantifier content type. This Nutil Quantifier content type can then be an input format for the MeshViewer Software (version x).

The other input should be defined as MeshView content type. Here it needs to be clarified what the valid structures are to be directly copy/past. It might be that we need to define one MeshView content type for CSV, for TSV and potentiall TXT. @apdavison what do you think?

MaaikevS: Ok, if we can cover the json with the Nutil content type, then it is fine with me. We may want to clean up those issues then, since it is very confusing as it is.

Regarding the second option for csv,tsv etc. I would not make a content type for this as you would not really use it as such, since the file itself cannot be uploaded to Meshview. 
I have extracted the x,y,z coordinates of the json file and stored them in a csv format, but the content type is just csv. The whole point it that that csv file is an open format and not linked to meshview

lzehl: I agree with cleaning up the issues. @Majpuc could you help here?

For the MeshView: this is not just a csv because it assumes a certain structure of the content. Nonetheless I need to discuss if copy/paste input qualifies as being registered as ContentTypes (@apdavison & @jagru20 your feedback is needed for this).

For now and for your purpose the Nutil Content Type is more important I guess?

MaaikevS: @lzehl I now changed meshview to nutil quantifier. Hope this is ok for now.
Could you have a look and merge if you are happy with it like this? 
Many thanks!

lzehl (approved):

MaaikevS:
Thanks @lzehl

PR #339, opened 2022-08-09
Create ISSN.schema.tpl.json

### LINKED ISSUE #338: Add ISSN to list of digital identifiers
needed by the "publications" module, to identify journals

https://www.issn.org
https://en.wikipedia.org/wiki/International_Standard_Serial_Number

The pattern should be something like `^[0-9]{4}-[0-9]{3}[0-9xX]$`

lzehl:
@apdavison I think the last part has to be a capital X if it is not a number. Could you double check this?


apdavison (approved):
This looks great. Formally it seems that an upper-case "X" is required. The regex given on in the Wikipedia page also allows a lower-case "x", I guess to make it case-insensitive, but I think it's fine to stick to upper-case only.

Issue #345, opened 2022-08-31
Registration of Data Use Agreements as licenses?


lzehl:
https://data.donders.ru.nl/doc/dua/RU-DI-HD-1.0.html?1

@apdavison and @UlrikeS91 specifically for externally published data, should we register DUAs as licenses (similar to the HDG EBRAINS DUAs we registered)?

apdavison:
for clarity, I suggest adding a new schema for DUAs

UlrikeS91:
A separate schema for DUAs is probably smarter. They do not have the same field as licenses and we already agreed when we added the EBRAINS DUAs that this is not a pretty fix but a good enough workaround for now. If we want to expand on this list, we should maybe fix it properly now.

lzehl:
@apdavison and @UlrikeS91 I agree with the both of you but we didn't do this for EBRAINS at the moment either because it entails a longer discussion (politically).

If you have a concrete suggestion that we could quickly implement (openMINDS is less of an issue here) and properly define a way of handling this in EBRAINS (display wise in the KG and politically wise because of missing licenses) please tell :) 

Otherwise I suggest we continue the workaround for now and get back to it later. 

What do you think?

UlrikeS91:
I assume that there won't be a quick solution, but we should raise this issue anyway. Until then, if you need the other DUAs urgently, I would agree that we stick to the workaround for now.

lzehl:
we can come back to finding a better solution later on in a separate issue. thanks for your quick feedback and support with this one.

PR #347, opened 2022-09-09
Create DondersInstitute-DUA-RU-DI-HD-1-0.jsonld

lzehl:
related to #345 

NOTE: the link to the DUA of the Donder's Institute is currently not working (there seems to be an error on their side). We should double check this later and correct the link if necessary. (merging the PR is urgent for a case in EBRAINS)


UlrikeS91 (approved):
I just removed the "?2" from the legal code. The page is still not working but this addition seems to have no function.

PR #362, opened 2022-11-07
ServiceLink from ModelVersion

apdavison:
The reasons for adding this rather than just using a ServiceLink from a File containing the model code is that:

1. ModelVersion only has the "repository" property, which often points to a URL outside EBRAINS (e.g. Github, ModelDB), for which individual Files are not indexed.
2. The ServiceLink URL can depend on multiple files, e.g. the inputData as well as the code.

### DIFF
--- a/schemas/data/serviceLink.schema.tpl.json
+++ b/schemas/data/serviceLink.schema.tpl.json
@@ -7,11 +7,12 @@
   ],
   "properties": {
     "dataLocation": {
-      "_instruction": "Add the file, file bundle, or atlas annotatation with the data that are linked to the specified service.",
+      "_instruction": "Add the file, file bundle, atlas annotatation, or other entity with the data that are linked to the specified service.",
       "_linkedTypes": [
         "https://openminds.ebrains.eu/core/File",
         "https://openminds.ebrains.eu/core/FileArchive",
         "https://openminds.ebrains.eu/core/FileBundle",
+        "https://openminds.ebrains.eu/core/ModelVersion",
         "https://openminds.ebrains.eu/sands/ParcellationEntityVersion",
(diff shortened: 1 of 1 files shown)


lzehl:
@apdavison I'm not sure I want to support this for the following reasons: 
+ we can index external file repositories, this is just not automatically done within EBRAINS.
+ you can link multiple files using file bundles 
+ we want to be explicit here because people tend to link generic concepts data to a service although only a portion of the concept / data is actually used by the service 

If you agree with my concerns I suggest to add FileRepository to the list instead of ModelVersion. That way it is at least a generic solution for all research products.

apdavison:
@lzehl I guess in my initial description I gave only practical reasons for this change, not conceptual ones. Conceptually, it is cleaner/more explicit in some cases to link ModelVersions to services, since the service is for working with a model, which just happens to be represented by some files. I think ModelVersion and SoftwareVersion are quite different from DatasetVersion in this regard, since both of the former usually represent singular items, whereas dataset versions are almost always collections, in which the individual items are represented by files or file bundles.

From a practical point of view, the service configuration may require certain metadata which come from the ModelVersion and are not present in the files. I'm thinking specifically of cases where multiple models are expressed by the same code, just with different parameterizations. 

We could work around this by creating a file bundle for each model version (generic code + specific parameter file), but this is a rather heavy workaround, since it would require the KG team to implement systematic indexing of Github and ModelDB-based repositories.

I'm keen to get this issue resolved soon, as it is the last remaining blocker to complete the migration of the Model Catalog/Model Validation service to KG v3, before KG v2 is turned off.

lzehl (approved):

lzehl:
@apdavison I understand. Then let's do it :)

Issue #353, opened 2022-09-13
adding 'performedBy' for ProtocolExecution or Activity


lzehl:
@apdavison @Peyman-N you should review this issue critically to see if this fits also for extensions

We (@UlrikeS91 and me) are thinking to add an optional 'performedBy' / 'agent' property to the ProtocolExecution.
I know that you typically ask for this info also for the extensions. Should we therefore add it rather to Activity (so that all extending schemas inherit the field)?

Peyman-N:
I prefer to have it in activity so the extensions can inherit it too. Originally there was a property named `agent` in the ephys and MRI. Beside that, this is an integral part of provenance information that we should capture. 
@apdavison what do you think?

apdavison:
To be honest, I thought it was already in Activity!

lzehl:
@Peyman-N and @apdavison :smile: then this is a clear GO for adding it to Activity. Which property name do you prefer ? ( aslo @UlrikeS91 ) I don't have a strong opinion.

apdavison:
In computation, we have the field "startedBy", since computations are assumed to be automated ones. W3C PROV uses "wasAssociatedWith" for the direct link, together with an optional "[qualifiedAssociation](https://www.w3.org/TR/prov-o/#narrative-example-qualified-4)" to specify the role taken by the agent. Simply "agent" would also be ok.

lzehl:
So if we add this now to Activity then it needs to work for all extended schemas. 

If you use "startedBy" in computation maybe "performedBy" (linking to -> SoftwareAgent, Person) would be a good option for all? 
"agent" might not be so intuitive as "performedBy".

Independent of the property name: we could also point to Contributing instead of Person to include the role that person played in the performance. We potentially have to reconsider though for this a bit the Contributing schema (name, and associated instances for roles). I'm open to make these changes since I was anyway not so happy currently with the schema and it is barely used as far as I know.

@apdavison @Peyman-N @UlrikeS91 what do you think?

UlrikeS91:
> So if we add this now to Activity then it needs to work for all extended schemas.
> 
> If you use "startedBy" in computation maybe "performedBy" (linking to -> SoftwareAgent, Person) would be a good option for all? "agent" might not be so intuitive as "performedBy".

I agree that "performedBy" is more intuitive and should fit all the examples that were mentioned here. 

[...]

PR #366, opened 2022-11-08
Update activity.schema.tpl.json

lzehl:
related to #353 @Peyman-N @apdavison @UlrikeS91 please all have a look and merge if all of you are happy. 

@apdavison & @Peyman-N this update might required corrections in the schemas that extend Activity (this might be called "startedBy" or something similar; for sure in computation extension). It might also require an update of EBRAINS registered instances (if you already registered agents for instances in computation activities).

### DIFF
--- a/schemas/research/activity.schema.tpl.json
+++ b/schemas/research/activity.schema.tpl.json
@@ -4,6 +4,15 @@
     "output"
   ],
   "properties": {
+    "customPropertySet": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "If any parameters are not covered in the standardized properties of the respective activity, it is possible to add here user-defined parameters grouped in context-specific sets.",
+      "_embeddedTypes": [
+        "https://openminds.ebrains.eu/core/CustomPropertySet"
+      ]
+    },
     "description": {
       "type": "string",
       "_instruction": "Enter a description of this activity."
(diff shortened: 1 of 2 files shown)


UlrikeS91 (approved):

UlrikeS91:
I'm happy with the PR. @apdavison and/or @Peyman-N should also review and can merge it if they are happy.

schemas/research/activity.schema.tpl.json
```diff
@@ -40,13 +49,14 @@
...
       "type": "array",
       "minItems": 1,
       "uniqueItems": true,
-      "_instruction": "If any parameters are not covered in the standardized properties of the respective activity, it is possible to add here user-defined parameters grouped in context-specific sets.",
-      "_embeddedTypes": [
-        "https://openminds.ebrains.eu/core/CustomPropertySet"
+      "_instruction": "Add all agents (persons or software agent) that performed this activity.",
+      "_linkedTypes": [
+        "https://openminds.ebrains.eu/core/Person",
+        "https://openminds.ebrains.eu/computation/SoftwareAgent"
```
apdavison: I know you've said, @lzehl, that you're comfortable with having links from core to other extensions, but I am not, I think it will hinder the evolution of openMINDS, and so I would prefer to avoid this going forwards and preserve a strict dependency tree for extensions.

In this case, how about defining a new category, "agent"?

lzehl: @apdavison we can introduce an agent category. It is an elegant way to "hide" core-extension linkages when they are not available.

But to clarify : I'm not comfortable with having links from core to other extensions. I just think sometimes it may not avoidable and as long as core can still be used without the extension (e.g. meaning the linkage to an extension is just an additional option to one present in core) I can live it.

lzehl: @apdavison "agent" category is added. see also https://github.com/HumanBrainProject/openMINDS_computation/pull/29

apdavison (approved):

Issue #360, opened 2022-10-04
request: add "consortium" as legal person 


lzehl:
There is a request from @maciej-jedynak to add a "consortium" schema as legal person

Definition: 
A consortium is an association of two or more individuals, companies, organizations or governments with the objective of participating in a common activity or pooling their resources for achieving a common goal. (https://en.wikipedia.org/wiki/Consortium)

I was thinking of making a simple schema: 
+ shortName
+ fullName
+ members
+ homepage
+ contactInformation

@apdavison @olinux @UlrikeS91 @bweyers what do you think?

apdavison:
ok with me

UlrikeS91:
Same here. Sounds useful, no objections 🙂

lzehl:
@olinux & @bweyers I will implement this request then. @olinux once it's done the EBRAINS KG (Editor & Search) needs an update I assume.

PR #367, opened 2022-11-08
Create consortium.schema.tpl.json

lzehl:
related to #360 (second try to do this PR :see_no_evil: )

### DIFF
--- a/schemas/actors/affiliation.schema.tpl.json
+++ b/schemas/actors/affiliation.schema.tpl.json
@@ -18,10 +18,11 @@
         "type": "string",
         "_instruction": "Enter the end date of this affiliation. Leave blank if the affiliation is still current."
      },
-     "organization": {
-       "_instruction": "Add organization to which a person is or was affiliated.",
+     "memberOf": {
+       "_instruction": "Add the organization or consortium to which a person is or was affiliated.",
        "_linkedTypes": [
-         "https://openminds.ebrains.eu/core/Organization"
+         "https://openminds.ebrains.eu/core/Organization",
+         "https://openminds.ebrains.eu/core/Consortium"
        ]
      }
    }
--- a/schemas/actors/affiliation.schema.tpl.json
+++ b/schemas/actors/affiliation.schema.tpl.json
@@ -3,27 +3,27 @@
   "required": [
     "organization"
   ],
-   "properties": {
-     "startDate": {
-        "_formats": [
-          "date"
-        ],
-        "type": "string",
-        "_instruction": "Enter the start date of this affiliation."
-     },
-     "endDate": {
-        "_formats": [
-          "date"
-        ],
-        "type": "string",
-        "_instruction": "Enter the end date of this affiliation. Leave blank if the affiliation is still current."
-     },
-     "memberOf": {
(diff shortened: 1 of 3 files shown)


jcolomb:
is this supposed to be an author of the dataset?

in jatsxml, collaboration group can have an affiliation and can have no member described.

lzehl:
@jcolomb yes. It came in as a request. So you would add an affiliation (optional) and keep members as optional? 
Would that not be then an Organization (all members with same affiliation) compared to a Consortium (members with different affiliations)? Or differently asked what is the affiliation of a Consortium with members with different affiliations? And a Consortium should always have members no? They might not be known though (why keeping it option makes sense).

I'm asking because we can currently have as "contributors" (authors, developers) Organizations and Persons. Consortium would be the third new option. Not sure to which the JATS schema(s) would relate best. Could you clarify?

jcolomb:
jats only has one entry for organisation and consortiums (called "collaborative group" I think), therefore they pack probably both types of entities under the same schema.

the issue with members is that it may change over time, the groups were not well described yet, especially in terms of what happens when the member list change.

lzehl:
@jcolomb the changing list of members of a consortium does not matter considering authorships I think. The legal person is here the consortium itself (independent of it's members). The members are just an additional information which indeed can change over time. 

@apdavison what are your thoughts here? do you have an idea if we could track such changes though over time for a consortium?

apdavison:
I would use the affiliation mechanism

lzehl:
@apdavison meaning you would not add the properties 'members' to a Consortium schema but rather point from Person to a Consortium via Affiliation ?

apdavison:
from Person and from Organization, yes. Affiliation has start date and end date, so allows membership to evolve over time.

lzehl:
@apdavison I've deleted the 'member' property from Consortium now, changed the property 'organization' in Affiliation schema to 'memberOf' and added Consortium there as linked type besides Organization.

What should we do with Organization. This schema currently does not have a link to Affiliation (which I think also does not make sense). Should I add there directly the property 'memberOf' with linked type to Consortium?

apdavison:
I think that adding Affiliation to Organization makes perfect sense, e.g. FZJ (organization) is affiliated to HBP (consortium).

lzehl:
@apdavison I don't think this is true... but maybe I have a different understanding of affiliation...

lzehl:
@apdavison I've added now affiliation to organizations as well

schemas/actors/affiliation.schema.tpl.json
```diff
@@ -18,10 +18,11 @@
         "type": "string",
         "_instruction": "Enter the end date of this affiliation. Leave blank if the affiliation is still current."
      },
-     "organization": {
-       "_instruction": "Add organization to which a person is or was affiliated.",
+     "memberOf": {
+       "_instruction": "Add the organization or consortium to which a person is or was affiliated.",
```
apdavison: or the consortium to which an organization is affiliated

lzehl: done

apdavison (approved):

Issue #368, opened 2022-11-28
adding other user IDs to Person schema


lzehl:
@apdavison you raised this issue within EBRAINS but I think it belongs here. 

Idea: besides defining the globally unique digital identifiers of a Person, allow a Person to specify user IDs for different accounts of interest (e.g. EBRAINS IAM, but also GitHub, GitLab, etc.)

Potential solution: 
- add 'accounts' (or ?) property to Person schema that links to (0-N) new UserAccount schema
- UserAccount schema properties:
  - accountType (-> link to UserAccountType ? in controlledTerms; terms: GitHub, GitLab, EBRAINS, etc OR would that link to an Organization?)
  - userName (string)

lzehl:
@HumanBrainProject/openminds-developers please have look and make suggestions

apdavison:
for account type, I would use the WebService schema, which I originally proposed for openMINDS_computation but I now think should go in core (see https://github.com/HumanBrainProject/openMINDS_computation/pull/18#issuecomment-1327169079)

lzehl:
@apdavison that is a good idea. But I do have some questions then about how you want to register a WebService. Because I don't see the current suggestion fit for GitHub or GitLab (but maybe I misunderstand it). What would be the version in these cases and who are the developers?

apdavison:
For GitHub, the developer would be the Github organisation, version "v1" or "n/a"

For self-hosted Gitlab, the developer would be GitLab Inc, I guess, and then the version number would reflect whichever version of the Gitlab software has been deployed.

lzehl:
@apdavison could you define the GitLab of EBRAINS as example?

Would that not also mean that we should make Web Services controlled instances of openMINDS? (since they would be valid for typically many cases) New ones should then be requested through openMINDS GitHub.

apdavison:
Something like this:
```
{
    "@id": "...",
    "@type": [
        "https://openminds.ebrains.eu/core/WebService"
    ],
    "https://openminds.ebrains.eu/vocab/fullName": "EBRAINS GitLab",
    "https://openminds.ebrains.eu/vocab/shortName": "EBRAINS GitLab",
    "https://openminds.ebrains.eu/vocab/hasVersion": [
      {
        "@id": "... (id of WebServiceVersion)"
      }
    ],
    "https://openminds.ebrains.eu/vocab/homepage": {
      "@id": "... (id of URL pointing to https://gitlab.ebrains.eu/api/v4)"
[...]

PR #371, opened 2022-12-06
Create accountInformation.schema.tpl.json for Person

lzehl:
related to #368

### DIFF
--- a/schemas/actors/accountInformation.schema.tpl.json
+++ b/schemas/actors/accountInformation.schema.tpl.json
@@ -6,7 +6,7 @@
   ],
    "properties": {
      "service": {
-       "_instruction": "Add the web service (version) of this account.",
+       "_instruction": "Add the web service of this account.",
        "_linkedTypes": [
          "https://openminds.ebrains.eu/core/WebService"
        ]
--- a/schemas/actors/accountInformation.schema.tpl.json
+++ b/schemas/actors/accountInformation.schema.tpl.json
@@ -1,5 +1,5 @@
 {
-  "_type": "https://openminds.ebrains.eu/core/Account",
+  "_type": "https://openminds.ebrains.eu/core/AccountInformation",
   "required": [
     "service",
     "userName"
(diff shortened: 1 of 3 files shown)


schemas/actors/accountInformation.schema.tpl.json
```diff
@@ -0,0 +1,19 @@
+{
+  "_type": "https://openminds.ebrains.eu/core/AccountInformation",
+  "required": [
+    "service",
+    "userName"
+  ],
+   "properties": {
+     "service": {
+       "_instruction": "Add the web service (version) of this account.",
```
apdavison: the instruction mentions the web service version, but the linked type is WebService. I think the latter is more appropriate, so the instruction should be changed.

schemas/actors/person.schema.tpl.json
```diff
@@ -7,6 +7,15 @@
     "givenName"
   ],
   "properties": {
+    "accountInformation": {
```
apdavison: I think "associatedAccount" might be a better term

schemas/actors/person.schema.tpl.json
```diff
@@ -7,6 +7,15 @@
...
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add one or several account information for this person.",
```
apdavison: Maybe "Add information about web service accounts held by this person."

schemas/actors/accountInformation.schema.tpl.json
lzehl: I agree. sorry this was a leftover when I was thinking to link both...

lzehl:
@apdavison should the schema name remain AccountInformation ? or should that be change to something else ?

apdavison:
@lzehl yes, I think AccountInformation is accurate.

apdavison (approved):

lzehl:
@olinux this merge requires an update in the EBRAINS KG we should discuss.

PR #373, opened 2022-12-20
Add optional expected content type to URL

apdavison:
This would be a very useful property for certain URLs, and would help avoid some ugly workarounds.


lzehl:
@apdavison note that this would be best followed by one or two ContentType instances for URLs

Issue #335, opened 2022-07-14
request to add identifiers.org IDs as option to DSV digitalIdentifier


lzehl:
Example for identifiers.org usage in Neurovault: https://identifiers.org/neurovault.collection:XXXX

@olinux I could use your help with this.

lzehl:
handled now in #372

PR #378, opened 2023-01-20
Create IdentifiersDotOrgID.schema.tpl.json

lzehl:
related to #335 

@olinux could you double check especially the regex in consideration of the [identification scheme](https://docs.identifiers.org/articles/docs/identification_scheme.html) of Identifiers.org ?
If it is okay I would add it to the possible list of digital identifiers for DSVs


lzehl:
due to the clean up I'll redo this PR

PR #379, opened 2023-01-24
Add RRID as an option for SoftwareVersion

apdavison:


lzehl (commented):
@apdavison we can add this but note that RRIDs are not necessarily assigned to versions but general constructs (version independent representations). Could you please add RRID as well to the Software schema?

apdavison:
that's a good point @lzehl

UlrikeS91:
I agree with this being part of both and took the liberty to update the software schema accordingly to speed up the process a bit 😉

lzehl (approved):

PR #385, opened 2023-02-10
Clean up: experimental activity

UlrikeS91:
MINOR changes:
- improved instructions

MAJOR changes:
- removed `_type` from the schema because concept schemas do not have a type and (as far as I am aware) can only be extended if they do not have a type themselves, correct @lzehl? This schema is extended a lot of times (e.g., specimenPrep/devicePlacement)

SPECIAL ATTENTION:
none

### DIFF
--- a/schemas/research/experimentalActivity.schema.tpl.json
+++ b/schemas/research/experimentalActivity.schema.tpl.json
@@ -1,5 +1,4 @@
 {
-  "_type": "https://openminds.ebrains.eu/core/ExperimentalActivity",
   "_extends": "research/activity.schema.tpl.json",
   "required": [
     "isPartOf",
@@ -13,7 +12,7 @@
       ]
     },
     "preparationDesign": {
-      "_instruction": "Add the initial preparation type for this experimental activity.",
+      "_instruction": "Add the initial preparation type for this activity.",
       "_linkedTypes": [
         "https://openminds.ebrains.eu/controlledTerms/PreparationType"
       ]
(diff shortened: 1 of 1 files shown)


lzehl (approved):
this is indeed only used as extension. it is a double extension (activity -> experimentalActivity -> XX) @olinux we need to make sure that the pipeline is doing this correctly. Currently not all of the extensions are working (I think the pipeline always fails if the concept schema is located in a different repo than the extending schemas)

lzehl:
@apdavison @Peyman-N feel free to merge when you are also happy with the changes

schemas/research/experimentalActivity.schema.tpl.json
```diff
@@ -13,7 +12,7 @@
       ]
     },
     "preparationDesign": {
-      "_instruction": "Add the initial preparation type for this experimental activity.",
+      "_instruction": "Add the initial preparation type for this activity.",
```
Peyman-N, line 15: why delete `experimental` from only this property?
I suggest either using `experimental activity` or `activity` for this schema and not alternate.

lzehl: agreed. I've removed "experimental" from all instructions

lzehl (approved):

lzehl:
@Peyman-N or @apdavison or @aeidi89 feel free to merge if you also approve now

Peyman-N (approved):

PR #387, opened 2023-02-10
Clean up: research product version

UlrikeS91:
MINOR changes:
- improved instructions
- establish correct order within the properties

MAJOR changes:
none

SPECIAL ATTENTION:
- removed maxLength for both description & shortName because I don't believe that we adhered to this so far and I don't see a good reason to have it here anymore
- removed the notion that no references are allow from the instructions of the description

CHANGELOG:
- `homepage` asks for IRI now instead of linking to `URL`
- `fullDocumentation` links to `WebResource` instead of `URL`

- adjustment of relatedPublication to fit the use for `livePaperVersion`:
     - add `ISSN` as allowed value
     - adjusted instructions to include the use of this property from the live paper version (original instruction there: "Add an identifier for the book, chapter, article, or preprint for which this live paper provides supplementary material.")

### DIFF
--- a/schemas/products/researchProductVersion.schema.tpl.json
+++ b/schemas/products/researchProductVersion.schema.tpl.json
@@ -18,7 +18,7 @@
       ]
     },
     "copyright": {
-      "_instruction": "Add the copyright information of this research product version.",
+      "_instruction": "Enter the copyright information of this research product version.",
       "_embeddedTypes": [
         "https://openminds.ebrains.eu/core/Copyright"
       ]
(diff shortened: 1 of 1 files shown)


schemas/products/researchProductVersion.schema.tpl.json
```diff
@@ -69,11 +68,11 @@
       "_instruction": "Enter the preferred citation text for this research product version. Leave blank if citation text can be extracted from the assigned digital identifier."
     },
     "keyword": {
-      "type": "array",
-      "maxItems": 5,
+      "type": "array",      
       "minItems": 1, 
+      "maxItems": 5,
```
lzehl: please remove this as well. if remove other restrictions we should not introduce one here. Each system (e.g. EBRAINS KG can decide this for themselves)

lzehl: adapt then again the instructions as well

lzehl (changes requested):

schemas/products/researchProductVersion.schema.tpl.json
UlrikeS91: Agreed and updated 🙂

schemas/products/researchProductVersion.schema.tpl.json
```diff
@@ -90,11 +90,12 @@
       "type": "array",
       "minItems": 1,
       "uniqueItems": true,
-      "_instruction": "Add all further publications besides the full documentation that provide the original context for the production of this research product version (e.g., an original research article).",
+      "_instruction": "Add all further publications besides the full documentation that provide the original context for the production of this research product version (e.g., an original research article) or receive supplementary material by this research product version.",
```
lzehl: I do not understand the addition. And "supplementary material" would include also learning resources which is explicitly not what is meant here.

lzehl (changes requested):

schemas/products/researchProductVersion.schema.tpl.json
```diff
@@ -91,19 +90,20 @@
...
       ]
     },
     "releaseDate": {
+      "type": "string",
       "_formats": [
         "date"
       ],
-      "type": "string",
-      "_instruction": "Enter the date (actual or intended) of the first broadcast/publication of this research product version."
+      "_instruction": "Enter the date (actual or intended) on which this research product version was first release, formatted as '2023-02-07'."
```
aeidi89: Suggest changing '2023-02-07' to 'YYYY-MM-DD'

lzehl (approved):
@aeidi89 I've updated the schema with your issue and adapted the instructions a bit of the "relatedPublication" property. Please double check and feel free to merge if you approve the changes.

Peyman-N (approved):

PR #389, opened 2023-02-10
Clean up: specimen set

UlrikeS91:
MINOR changes:
- improved instructions

MAJOR changes:
none

SPECIAL ATTENTION:
- removed `quantity` from concept schema, but property won't disappear; each specimen set gets a more specific property (`numberOfXY`) which askes for the same information

### DIFF
--- a/schemas/research/specimenSet.schema.tpl.json
+++ b/schemas/research/specimenSet.schema.tpl.json
@@ -8,7 +8,7 @@
   "properties": {
     "additionalRemarks": {
       "type": "string",
-      "_instruction": "Enter additional remarks about the specimen set."
+      "_instruction": "Enter any additional remarks concering this specimen set."
     },
     "biologicalSex": {
       "type": "array",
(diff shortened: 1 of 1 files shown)


lzehl (approved):

lzehl:
@apdavison @Peyman-N could one of you confirm in addition (also to make sure that you are aware of this change)

schemas/research/specimenSet.schema.tpl.json
```diff
@@ -21,21 +21,17 @@
...
-      "type": "integer",
-      "_instruction": "Enter the number of specimen that belong to this set."
+      "_instruction": "Enter a lookup label for this specimen set that may help you to find this instance more easily."
     },
     "species": {
       "type": "array",
       "minItems": 1,
       "uniqueItems": true,
-      "_instruction": "Add the species and/or the strain (a sub-type of a genetic variant of species) of all specimen in this set.",
+      "_instruction": "Add the species and/or strain (a sub-type of a genetic variant of species) of all specimen in this set.",
```
Peyman-N, line 34: Just a suggestion maybe `a sub-type of a species with a unique genetic profile` is a little bit more specific. 
`Add the species and/or strain (a sub-type of a species with a unique genetic profile) of all specimen in this set.` 
I am not a biologist so I am not sure :sweat_smile:

Peyman-N (commented):
Just another remark I think "specimen" should be pluralized as "specimens" since it refers to multiple specimens. 
`      "_instruction": "Add the biological sex of all specimens in this set.",`

schemas/research/specimenSet.schema.tpl.json
aeidi89: Having worked a bit with animals, for me the description of strain is not necessary. However, if it is helpful I would mind keeping it. Good to include all :)

lzehl: I have the same opinion here. For me a species can be a strain or simple species definition. I would not need the additional information in brackets. @Peyman-N I'm not sure that your suggestion makes it more specific... do you insist on an update here?

Peyman-N: I am okay with it, so i would merge it

Peyman-N (approved):

PR #390, opened 2023-02-10
Clean up: specimen state

UlrikeS91:
MINOR changes:
- improved instructions

MAJOR changes:
none

SPECIAL ATTENTION:
none

### DIFF
--- a/schemas/research/specimenState.schema.tpl.json
+++ b/schemas/research/specimenState.schema.tpl.json
@@ -14,6 +14,10 @@
         "https://openminds.ebrains.eu/core/QuantitativeValueRange"
       ]
     },
+    "internalIdentifier": {
+      "type": "string",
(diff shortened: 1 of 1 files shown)


schemas/research/specimenState.schema.tpl.json
```diff
@@ -5,38 +5,38 @@
   "properties": {
```
lzehl, line 5: @UlrikeS91 this has nothing to do with the cleanup but it came up during a curation in EBRAINS now. We will also need an internalIdentifier for specimenStates (optional). Could you add it?

lzehl (changes requested):

schemas/research/specimenState.schema.tpl.json
UlrikeS91: done

lzehl (approved):

PR #393, opened 2023-02-10
Clean up: comment

UlrikeS91:
MINOR changes:
- improved instructions
- renamed `subject` to `about` to reduce the list of property names and to avoid potential confusion in regards to the specimen "subject" which we use more commonly (Note: the term 'subject' is of course not wrong here, but given our context we should try to avoid this additional meaning of the term) 
- renamed `content` to `comment` to make this more straight-forward/specific and to avoid potential confusion in regards to the existing property `contentDescription` where we want a description of the content of something (e.g., a file); here we want a user to enter the text of the comment that they would like to make, so we can be specific about that
 
MAJOR changes:
none

SPECIAL ATTENTION:
none

### DIFF
--- a/schemas/miscellaneous/comment.schema.tpl.json
+++ b/schemas/miscellaneous/comment.schema.tpl.json
@@ -1,26 +1,26 @@
 {
   "_type": "https://openminds.ebrains.eu/core/Comment",
   "required": [
-    "subject",
-    "content",
+    "about",
+    "comment",
     "commenter",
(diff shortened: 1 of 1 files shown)


lzehl (approved):

lzehl:
@apdavison or @Peyman-N one of you please double check and merge if you are happy.

Peyman-N (approved):

PR #395, opened 2023-02-10
Clean up: consortium

UlrikeS91:
MINOR changes:
- improved instructions

MAJOR changes:
none

SPECIAL ATTENTION:
none

CHANGELOG:
- `homepage` asks for IRI now instead of linking to `URL`

### DIFF
--- a/schemas/actors/consortium.schema.tpl.json
+++ b/schemas/actors/consortium.schema.tpl.json
@@ -22,7 +22,7 @@
       "_formats": [
         "iri"
       ],
-      "_instruction": "Enter the internationalized resource identifiers (IRIs) to the homepage of this consortium."
+      "_instruction": "Enter the internationalized resource identifier (IRI) to the homepage of this consortium."
     },
     "shortName": {
       "type": "string",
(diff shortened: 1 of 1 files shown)


Peyman-N (approved):

Peyman-N:
Just a note, won't it be more appropriate to have a consortium members property? As by definition consortium is association of multiple organisations or peoples. 

P.S. Maybe I should move this to a issue so we can deal with it latter.

UlrikeS91:
> Just a note, won't it be more appropriate to have a consortium members property? As by definition consortium is association of multiple organisations or peoples.
> 
> P.S. Maybe I should move this to a issue so we can deal with it latter.

This has actually been part of the schema in its original draft (see original issue https://github.com/HumanBrainProject/openMINDS_core/issues/360 and PR https://github.com/HumanBrainProject/openMINDS_core/pull/367). The issue with the members list has been discussed in the PR. The rough summery is that the maintenance of the members of a consortium will be too much work, and @apdavison suggested to rather handle this relationship via "affiliations".

lzehl (approved):

PR #399, opened 2023-02-13
Clean up: contribution 

UlrikeS91:
MINOR changes:
- improved instructions
- renamed `contributionType` to `type` to reduce list of property names and because we decided to use type when the specific context is explicit (a contribution schema would expect a contribution type, so we don't need the property name to specify this in such detail)

MAJOR changes:
none

SPECIAL ATTENTION:
none


lzehl (approved):

lzehl:
@apdavison or @Peyman-N feel free to merge when you are also happy.

Peyman-N (approved):

PR #400, opened 2023-02-13
Clean up: copyright

UlrikeS91:
MINOR changes:
- improved instructions

MAJOR changes:
none

SPECIAL ATTENTION:
none

### DIFF
--- a/schemas/data/copyright.schema.tpl.json
+++ b/schemas/data/copyright.schema.tpl.json
@@ -9,7 +9,7 @@
       "type": "array",
       "minItems": 1,
       "uniqueItems": true,
-      "_instruction": "Add one or several persons or organisations that hold the copyright.",
+      "_instruction": "Add all parties that hold this copyright.",
       "_linkedCategories": [
          "legalPerson"
       ]
@@ -17,7 +17,7 @@
     "year": {
       "type": "string",
       "pattern": "([0-9]{4})",
-      "_instruction": "Enter the year during which the copyright was first asserted."
+      "_instruction": "Enter the year during which the copyright was first asserted follwoing the defined pattern (e.g., 2023)."
     }
   }
 }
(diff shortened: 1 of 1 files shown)


schemas/data/copyright.schema.tpl.json
```diff
@@ -9,15 +9,15 @@
...
+      "_instruction": "Add all parties that hold this copyright.",
       "_linkedCategories": [
          "legalPerson"
       ]
     },
     "year": {
       "type": "string",
       "pattern": "([0-9]{4})",
-      "_instruction": "Enter the year during which the copyright was first asserted."
+      "_instruction": "Enter the year during which the copyright was first asserted follwoing the defined pattern (e.g., 2023)."
```
lzehl: @apdavison this is something we need to overthink correct? in principle the copyright has to be renewed for each year which would make the link to this type an array and then this instruction should read something like this: "Enter the year during which the copyright was claimed following the defined pattern (e.g., 2023)."

@UlrikeS91 besides this issue there is a typo in "follwoing" :wink:

lzehl (changes requested):

schemas/data/copyright.schema.tpl.json
UlrikeS91: I'll wait for feedback form @apdavison before making any changes

apdavison: yes, an array would be a nice solution. The instructions should then be something like "Enter the year during which the copyright was first asserted and, optionally, later years during which updated versions were published."

lzehl: perfect. thanks for the feedback. I'll do the changes

lzehl (approved):
@apdavison I've updated the schema. Please double check and feel free to merge if you approve. Question: should the copyright on RPVs should also become an array? (in case the copyright holder changed for a published version?)

Peyman-N (approved):

PR #401, opened 2023-02-13
Clean up: custom property set

UlrikeS91:
MINOR changes:
- improved instructions
- renamed `definedIn` with `dataLocation`, which merges two existing properties which were used in approx. the same way; details of what data is found there or how the data is stored, should be stated in the instructions

MAJOR changes:
none

SPECIAL ATTENTION:
none

CHANGELOG:
- replaced link to technique for `relevantFor` with new category `technique` (which includes techniques, analysisTechniques and stimulationTechniques); see https://github.com/HumanBrainProject/openMINDS_controlledTerms/pull/393

### DIFF
--- a/schemas/research/customPropertySet.schema.tpl.json
+++ b/schemas/research/customPropertySet.schema.tpl.json
@@ -2,26 +2,26 @@
   "_type": "https://openminds.ebrains.eu/core/CustomPropertySet",
   "required": [
     "context",
-    "relevantFor",
-    "definedIn"
+    "dataLocation",    
+    "relevantFor"
   ],
   "properties": {
     "context": {
       "type": "string",
       "_instruction": "Enter the common context for this custom property set."
     },
+    "dataLocation": {
+      "_instruction": "Add the location of the data that define the custom property set for the given context (e.g., stored as file or other entities such as property-value lists).",
+      "_linkedTypes": [        
+        "https://openminds.ebrains.eu/core/Configuration",
+        "https://openminds.ebrains.eu/core/File",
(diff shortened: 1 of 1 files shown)


schemas/research/customPropertySet.schema.tpl.json
```diff
@@ -2,27 +2,27 @@
...
-    "definedIn"
+    "dataLocation",    
+    "relevantFor"
   ],
   "properties": {
     "context": {
       "type": "string",
       "_instruction": "Enter the common context for this custom property set."
     },
+    "dataLocation": {
```
lzehl, line 13: or we call this "about" ?? I think the meaning would fit better. @apdavison @Peyman-N we need your feedback here

lzehl (changes requested):

schemas/research/customPropertySet.schema.tpl.json
UlrikeS91: No, `about` will be wrong in this context. What we expect, is the location of the file (or other entities) that defines this set. We used `about` in schemas to define what the schema is about (e.g., a `learningResource` could be about a `dataset`). Here we are asking for the file (or other entity) that is about this set - not the other way around.

lzehl: okay with me.

lzehl (approved):

lzehl:
@Peyman-N or @apdavison feel free to merge if you also approve

Peyman-N (approved):

PR #421, opened 2023-02-13
Clean up: model version

UlrikeS91:
MINOR changes:
- improved instructions

MAJOR changes:
none

SPECIAL ATTENTION:
- renamed `format` to `usedFormat`: because format has been used to express the format of the type of schema, e.g. file has a format with instructions to add the content type OF the file, here the instructions ask for USED content types

CHANGELOG:
- `inputData` & `outputData` link to `WebResource` instead of `URL`

### DIFF
--- a/schemas/products/modelVersion.schema.tpl.json
+++ b/schemas/products/modelVersion.schema.tpl.json
@@ -2,8 +2,8 @@
   "_type": "https://openminds.ebrains.eu/core/ModelVersion",
   "_extends": "products/researchProductVersion.schema.tpl.json",
   "required": [
-    "license",
-    "format"
+    "format",
+    "license"
   ],
   "properties": {
     "developer": {
--- a/schemas/products/modelVersion.schema.tpl.json
+++ b/schemas/products/modelVersion.schema.tpl.json
@@ -70,11 +70,11 @@
         "https://openminds.ebrains.eu/core/WebResource"
       ]
     },    
-    "usedFormat": {
+    "format": {
       "type": "array",
       "minItems": 1,
       "uniqueItems": true,
-      "_instruction": "Add all used content types of this computational model version.",
+      "_instruction": "Add all content types of the files composing this computational model version.",
       "_linkedTypes": [
         "https://openminds.ebrains.eu/core/ContentType"
       ]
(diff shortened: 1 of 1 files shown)


apdavison (commented):
I'm not convinced about this distinction between "format" and "usedFormat".

Sometimes a research object/product is contained in a single file with a given content type. In that case I think "format" is more appropriate than "usedFormat".

Sometimes, a research object is distributed across multiple files, each of which has their own content type, but the group of files as a whole also has a content type (e.g. BIDS, SONATA). In this case, I think "format" is also the most appropriate.

Sometimes, a research object consists of multiple files, but the collection as a whole does not have its own content type. This is where "usedFormat" might be appropriate.

So overall, either we have both "format" and "usedFormat", or we have just a single property relating to content types. I think the former is too fiddly, so I suggest keeping just a single property, with the name "format".

lzehl (approved):

lzehl:
@apdavison or @Peyman-N feel free to merge if you are also happy with the changes

lzehl (commented):
@apdavison if this is only referring to the format of the files defining this computational model version then this should be defined in the FileRepository schema, not on the ModelVersion schema. We assumed that this property refers to something that is used by the ModelVersion not that represents the ModelVersion

lzehl:
@apdavison we need further input from you here

apdavison:
As I said above, sometimes it is the content type *of* the model, other times, where we don't have a content type for the model as such, it is the list of content types of the files that make up the model.

I think the best approach is to keep the property name `format` (and allow it to be an array).

lzehl:
@apdavison that would still mean that this is redundant information that is also given in the file repository no?

lzehl:
Maybe we should discuss this one in person. I'm not sure I understand it correctly.

lzehl (commented):
@apdavison I've changed it back to "format" and edited the instructions. Please double check. I would still like to discuss this one to understand your argument a bit better. Feel free to merge the PR though if you approve this PR now

apdavison:
@lzehl I've fixed a few things, please check.

lzehl (approved):
@apdavison thank you for realizing that I forgot to update the rest with changing the property name back :see_no_evil:

lzehl:
@apdavison can we merge?

PR #436, opened 2023-02-14
Clean up: setup

UlrikeS91:
MINOR changes:
- improved instructions
- fixed indentation
- renamed `setupType` to `type` to reduce list of property names and because we decided to use type when the specific context is explicit

MAJOR changes:
none

SPECIAL ATTENTION:
none

### DIFF
--- a/schemas/products/setup.schema.tpl.json
+++ b/schemas/products/setup.schema.tpl.json
@@ -13,7 +13,7 @@
       "type": "array",
       "minItems": 2,
       "uniqueItems": true,
-      "_instruction": "Add all components, including other setups, that are part in this setup. The setup must be composed of at least one device.",
+      "_instruction": "Add all components, including other setups, that are part in this setup. Note that a setup should not be only composed of software.",
       "_linkedCategories": [
         "setupComponent"
       ]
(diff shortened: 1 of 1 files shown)


Peyman-N (approved):

schemas/products/setup.schema.tpl.json
```diff
@@ -1,52 +1,52 @@
...
+    "components",    
+    "description",
+    "name"
+  ],
+  "properties": {    
+    "components": {
+      "type": "array",
+      "minItems": 2,
+      "uniqueItems": true,
+      "_instruction": "Add all components, including other setups, that are part in this setup. The setup must be composed of at least one device.",
```
lzehl: this seems like contradiction... minItems 2 but only one device is required according to the instructions...

lzehl (changes requested):

schemas/products/setup.schema.tpl.json
lzehl: I suggest to just delete the second sentence.

lzehl (changes requested):

lzehl (approved):
updated instructions of "components" to "Add all components, including other setups, that are part in this setup. Note that a setup should not be only composed of software."

schemas/products/setup.schema.tpl.json
lzehl: changed to "Add all components, including other setups, that are part in this setup. Note that a setup should not be only composed of software."

PR #439, opened 2023-02-14
Clean up: stock number

UlrikeS91:
MINOR changes:
- improved instructions

MAJOR changes:
none

SPECIAL ATTENTION:
- @lzehl, should this schema be moved to the new folder for `digitalIdentifier` or remain where it is?


lzehl (approved):

lzehl:
@UlrikeS91 not sure if they fit into "digital identifiers". Are stock numbers following any global pattern? Are they globally unique? I assume they are persistent. @apdavison I need your input for this decision. @tgbugs what is your thought here?

Peyman-N:
I don't believe that a stock number is a true identifier, as it can be modified based on the organization's requirements.
The identifier version of it would be ` Global Trade Item Number` or GTIN. 
Any how where have we used this schema?

Peyman-N (approved):

lzehl:
> I don't believe that a stock number is a true identifier, as it can be modified based on the organization's requirements. The identifier version of it would be ` Global Trade Item Number` or GTIN. Any how where have we used this schema?

we use it in the Strain schema; @UlrikeS91 could you investigate if we could ask for GTIN here instead or in addition? if so we could move this to the digitalIdentifier folder.

jcolomb:
I do not know for mice, but for flies, stock numbers can be a PID, if people use the Flybase version of the stock number (otherwise, you need both the number and the organisation using that number to identify the stock).
I do think but I am not sure , this can be quite similar using MGI numbers for mice
see https://www.informatics.jax.org/strain/MGI:3056279

lzehl:
@jcolomb thanks for your feedback on this one. I think then we should move this to our digital identifiers folder.

lzehl:
@Peyman-N please double check and merge if you approve again.

lzehl (approved):

lzehl:
or @aeidi89 please double check and feel free to merge if everything seems correct

Peyman-N (approved):

PR #441, opened 2023-02-14
Clean up: string property

UlrikeS91:
MINOR changes:
- improved instructions

MAJOR changes:
none

SPECIAL ATTENTION:
none

### DIFF
--- a/schemas/research/stringProperty.schema.tpl.json
+++ b/schemas/research/stringProperty.schema.tpl.json
@@ -11,7 +11,7 @@
     },
     "value": {
       "type": "string",
-      "_instruction": "Enter a text value for this property."
+      "_instruction": "Enter the text value that is described by this string property."
     }
   }
 }


schemas/research/stringProperty.schema.tpl.json
```diff
@@ -11,7 +11,7 @@
     },
     "value": {
       "type": "string",
-      "_instruction": "Enter a text value for this property."
+      "_instruction": "Enter the text value that is described by this string property."
```
lzehl, line 14: that seems wrong.... "Enter the text value of this string property."

lzehl (commented):
@apdavison or @Peyman-N I need your input here

schemas/research/stringProperty.schema.tpl.json
Peyman-N: No I think it is correct, but I am not sure.

Peyman-N (approved):

schemas/research/stringProperty.schema.tpl.json
Peyman-N: Either choice seems good enough, so I'm happy to approve it no matter which one you go with.

lzehl: for me "is described by" means that this is indirect (I'm describing something...) but here the text value itself is the entity. But maybe it's just me. leave it if you prefer it

lzehl (approved):

PR #454, opened 2023-02-14
Clean up: device usage

UlrikeS91:
MINOR changes:
none

MAJOR changes:
- introduction of new concept schema for device usages (in ephys and specimenPrep repo)

SPECIAL ATTENTION:
- specific device names are replaced by generic `device` as property name (e.g. electrode schema had `electrode` to state the electrode (device) that was used)
- former `additionalInformation` renamed to `metadataLocation` to avoid potential confusions with `additionalRemarks` 
- all 3 properties are newly introduced for some of the usage schemas (`SlicingDeviceUsage` in specimenPrep repo, which had none of them)

### DIFF
--- a/schemas/research/deviceUsage.schema.tpl.json
+++ b/schemas/research/deviceUsage.schema.tpl.json
@@ -1,0 +1,31 @@
+{
+  "_categories": [
+    "deviceUsage"
+  ],
+  "required": [
+    "device"
+  ],
+  "properties": { 
+    "metadataLocation": {    
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add all files or file bundles containing additional information about the usage of this device.",
+      "_linkedTypes": [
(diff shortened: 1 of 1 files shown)


Peyman-N:
@UlrikeS91 I thought we decided against `deviceUsage`? 
Anyhow the schema itself make lots of sense.

lzehl:
@Peyman-N what do you mean "we decided against 'deviceUsage'" ?  We are having specific device usage schemas. @UlrikeS91 just makes the suggestion to create a concept schema for the common properties of all the specific device usage schemas (to spear us some maintenance work). I think we never decided against a concept schema. We did decide against a generic device usage schema, though (but that is not the same). Or did I forget something about why a concept schema is not a good idea here?

Peyman-N:
@lzehl No no, you are wright. I think this consent schema seems good enough.

Peyman-N (approved):

lzehl (approved):

Issue #361, opened 2022-10-25
grouping research products under a user-defined topic in projects


lzehl:
I received a request for the possibility of grouping research products under a user-defined topic in projects and also if it would be possible to let the user decided in which order these groupings are displayed (the latter part actually is not really an openMINDS issue but a matter of how the data management system interprets openMINDS: @olinux that would affect the EBRAINS KG).

If projects are composed of many research products this request makes a lot of sense.
As small example (organization of a dataset list under a project):
- "Datasets containing raw data:"
  - dataset A
  - dataset B
  - ...
  - dataset Z
- "Datasets containing derived data of processing step 1:"
  - dataset A.1
  - dataset B.1
  - ...
  - dataset Z.1

I think this could be done with a small grouping schema with the properties:
- "grouping context" (string, free text)
- "group items" (obj, linked category: research products)

@olinux @apdavison @UlrikeS91 @bweyers I'd like to hear your thoughts / ideas for this issue / request. It would be nice to find a solution.

apdavison:
This might be a good use case for a Live Paper, which allows such groupings plus labels/commentary

PR #457, opened 2023-02-28
Create researchProductGroup.schema.tpl.json

lzehl:
related to #361 

@apdavison or @aeidi89 would you agree with the suggested schema? Nothing is set in stone yet. I just wanted to get this on it's way for the release

### DIFF
--- a/schemas/miscellaneous/researchProductGroup.schema.tpl.json
+++ b/schemas/miscellaneous/researchProductGroup.schema.tpl.json
@@ -1,0 +1,21 @@
+{
+  "_type": "https://openminds.ebrains.eu/core/ResearchProductGroup",
+  "required": [
+    "values"
+  ],
+  "properties": {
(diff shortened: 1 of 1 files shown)


schemas/miscellaneous/researchProductGroup.schema.tpl.json
```diff
@@ -0,0 +1,21 @@
+{
+  "_type": "https://openminds.ebrains.eu/core/ResearchProductGroup",
+  "required": [
+    "values"
```
apdavison: there's no property "values"

lzehl: :see_no_evil: of course...

lzehl: corrected! sorry for that stupid error

Issue #372, opened 2022-12-12
adding identifiers.org identifiers as digital identifiers


lzehl:
cf scheme:
https://docs.identifiers.org/articles/docs/identification_scheme.html

should definitely be an option for DSV digitial identifier (example neurovault). potentially also elsewhere? 
@apdavison @bweyers @UlrikeS91 @skoehnen @olinux @jagru20 please comment

UlrikeS91:
Draft in #378

PR #458, opened 2023-02-28
Create IdentifiersDotOrgID.schema.tpl.json

lzehl:
related to #372


apdavison (approved):
looks ok, although I've no idea if the regex is correct!

olinux (approved):
same here - didn't revalidate the regex, but other than that it looks ok

lzehl (commented):
@apdavison & @olinux thanks. I've tested the regex now against a couple of examples and it worked. Note that in principle each registry in Identifier.org builds on this regex to specify the pattern of the registry (e.g., zinc:ZINC?\d+). We need to see if these more detailed patterns have to registered with us as well...

Issue #463, opened 2023-03-14
could we fix the "née" structurally for persons for the release?


lzehl:
currently we integrate this into the string of the familyName (e.g., "Ruland (née Bradler)").
I would rather extract this into it's own property field (optional of course).

@apdavison @olinux @Peyman-N @aeidi89 what do you think?

lzehl:
**Note:** I've used the property name "birthName" now which is a bit imprecise, since this may be applied to the family name, the given name, or the entire name. I stated in the instructions though that we mean the family name here. Feel free to suggest a different property name / different approach to handle name changes. In EBRAINS they have use cases for family name changes due to marriages, they have no use case yet for a given name change in the KG (which is also possible).

lzehl:
@jcolomb how do you handle this in **jams**?

lzehl:
Based on the discussion with @aeidi89 we suggest to use "alternateFamilyName" and "alternateGivenName".

apdavison:
I agree with using "alternate...", since people change their names for many reasons.

jcolomb:
We do not have talked about this in jams, I am not aware of anything in jatsxml. orcid uses "Also known as" which would be full name alternative.

lzehl:
@jcolomb thanks for your feedback. ORCID matches roughly what schema.org is doing. I think we are then a bit more specific but close enough to running standards with "alternateFamilyName" and "alternateGivenName"

jcolomb:
Wondering if "alternateName" would not be simpler ( imagine if someone change their family and given name (at different time), would you need to compute all different possibilities to get all the alternate names ?)

lzehl:
Mh... @jcolomb I see your point, but keeping it together makes it difficult to track what changed and if for example a system should show the family name as "FAMILYNAME (aka ALTERNATEFAMILYNAME)" (which is wanted by users in EBRAINS)

lzehl:
done with #468

PR #464, opened 2023-03-14
Update person.schema.tpl.json with "birthName"

lzehl:
related to #463 

I can adapt the property name and instructions based on the provided feedback.

### DIFF
--- a/schemas/actors/person.schema.tpl.json
+++ b/schemas/actors/person.schema.tpl.json
@@ -17,6 +17,24 @@
         "https://openminds.ebrains.eu/core/Affiliation"
       ]
     },
+    "alternateFamilyName": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "minItems": 1,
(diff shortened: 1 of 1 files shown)


schemas/actors/person.schema.tpl.json
```diff
@@ -17,6 +17,24 @@
...
+      "items": {
+        "type": "string"
+      },
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Enter any other known given name of this person."
+    },
     "associatedAccount": {
       "type": "array",
       "minItems": 1,
```
aeidi89, line 40: Is there are reason for why there is no:

      "uniqueItems": true,
      "_instruction": "Enter any other known family name of this person."
      
      for the "associatedAccount"?

@lzehl

aeidi89 (commented):
Looks good, but I have a (newbie) comment that want to get checked out before approving.

aeidi89 (approved):
Fixed

PR #470, opened 2023-03-20
Add category "publicationReference"

apdavison:
as discussed in Slack, allowing linking ScholarlyArticle, Chaper, Book instances from openMINDS_publication module in addition to DOI, ISBN, etc


apdavison:
should be merged together with https://github.com/HumanBrainProject/openMINDS_publications/pull/18

lzehl (approved):

PR #488, opened 2023-08-30
Add a GenericIdentifier schema

apdavison:
cf https://github.com/HumanBrainProject/openMINDS/issues/51

### DIFF
--- a/schemas/digitalIdentifier/genericIdentifier.tpl.json
+++ b/schemas/digitalIdentifier/genericIdentifier.tpl.json
@@ -1,0 +1,23 @@
+{
+    "_type": "https://openminds.ebrains.eu/core/GenericIdentifier",
+    "required": [
+      "identifier",
+      "emitter"
+    ],
+    "properties": {
+      "identifier": {
+        "type": "string",
+        "_instruction": "Enter a persistent, unique identifier emitted by an organization."
+      },
+      "type": {
+        "type": "string",
+        "_instruction": "Enter the type of identifier, e.g. 'PubMed ID'."
+      },
+      "emitter": {
+        "_instruction": "Add the organization that governs and/or emits the identifier.",
+        "_linkedTypes": [
+          "https://openminds.ebrains.eu/core/Organization"
+        ]
+      }
+    }
+  }
--- a/schemas/digitalIdentifier/genericIdentifier.tpl.json
+++ b/schemas/digitalIdentifier/genericIdentifier.tpl.json
@@ -1,23 +1,21 @@
 {
-    "_type": "https://openminds.ebrains.eu/core/GenericIdentifier",
-    "required": [
-      "identifier",
-      "emitter"
-    ],
-    "properties": {
-      "identifier": {
-        "type": "string",
-        "_instruction": "Enter a persistent, unique identifier emitted by an organization."
-      },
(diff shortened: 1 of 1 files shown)


schemas/digitalIdentifier/genericIdentifier.tpl.json
```diff
@@ -0,0 +1,23 @@
...
+    "required": [
+      "identifier",
+      "emitter"
+    ],
+    "properties": {
+      "identifier": {
+        "type": "string",
+        "_instruction": "Enter a persistent, unique identifier emitted by an organization."
+      },
+      "type": {
```
lzehl: I think this should be rather called "name". I assume we want to leave this flexible ?

schemas/digitalIdentifier/genericIdentifier.tpl.json
```diff
@@ -0,0 +1,23 @@
+{
```
lzehl, line 1: formatting: 2 spaces / tab; newline missing at the end of the file; alphabetic order of properties

lzehl (changes requested):
Please double check in how far this issue is not already covered by the core schema for Identifiers.org (for Pubmed cf. https://registry.identifiers.org/registry/pubmed)

apdavison:
> Please double check in how far this issue is not already covered by the core schema for Identifiers.org (for Pubmed cf. https://registry.identifiers.org/registry/pubmed)

identifiers.org is limited in scope to life sciences. As an example, it doesn't include https://hal.science identifiers (as far as I can see) or other national science repositories.

schemas/digitalIdentifier/genericIdentifier.tpl.json
apdavison: "name" usually refers to instances.

schemas/digitalIdentifier/genericIdentifier.tpl.json
apdavison: done

lzehl:
@apdavison should we then not define "type" / "name" in the schema but just identifier and the emitter? I think we should push people to register official identifiers through typed schemas (because they can actually validate the format of the identifier).

I agree that the typing all of them is though a bit cumbersome. The clean alternative for this would be the following:

+ schema for identifierType: (with controlled instance library in openMINDS)
  - name (string)
  - regex pattern (string, regex)


+ schema for digitalIdentifier:
  - type (link to identifierType)
  - identifier
  - emitter (optional) [or would that go to the identifierType?] 

Research products would then only point to digitalIdentifier schema.
Unfortunately this approach would cause a lot of change to the running system.
It might be worth it though...

apdavison:
That would be a cleaner way to do it. I suggest we think about making that change for the next-but-one release. This schema is then a stop-gap, to provide the flexibility needed in https://github.com/HumanBrainProject/openMINDS/issues/51 but without defining a whole bunch of explicit identifier schemas.

The schema needs "type" or "kind", since a given Organization might emit more than one type of identifier.

lzehl:
@apdavison I agree with the plan to push this change to the next development round. (also because Oli would kill me if we do this now :see_no_evil: )

lzehl (approved):

PR #503, opened 2024-04-12
Typo in instructions

UlrikeS91:


lzehl (approved):
this is backwards compatible and could be lead to an update of v3.0 openMINDS

PR #504, opened 2024-04-15
Allow ISBN under fullDocumentation (RPV)

UlrikeS91:
This is needed for the atlas that I am registering at the moment. All information about the atlas is part of the book. So, the only ID that can be used in this case is the ISBN of the book.


lzehl (approved):
this is backwards compatible and can be updated in v3.0 openMINDS (in a sub-release)

PR #512, opened 2024-09-19
V5: Removal of namespace in schema files.

Raphael-Gazzotti:


lzehl (approved):
if the prior PR #484 goes through this should be integrated here before merge.

lzehl (approved):

PR #524, opened 2024-12-03
First data type schemas

### LINKED ISSUE #523: First schema suggestions for specific data types to be added in core
This was started in the [openMINDS_neuroimaging extension](https://github.com/openMetadataInitiative/openMINDS_neuroimaging/tree/dev/schemas/entity), but @lzehl and I would like to suggest to move them to core. The main reason is that any research product could profit from using them as (sort of) extensions to their associated files to provide more details about the data. It would be cumbersome to have to use multiple openMINDS extensions to find suitable data type schemas, the list of data types is not as extensive as e.g., devices where we concluded that we collect them in the most suitable openMINDS extension repos instead and it would also allow us to link to these specific data types as the data for specific schemas like e.g., coordinate frameworks in the upcoming SANDS update. The latter would be important in cases where the specific files may not be easy to share, but to still provide some metadata. This logic may also apply to other extensions as well, but should be discussed in a developers meeting. 

I have made a draft for some data types based on the draft for some of the data types that @Peyman-N has prepared in the neuroimaging extension. This draft includes the types needed to cover any kind of files used as reference data for atlases. 

### RasterGraphic
| property | required? | count | value | comment |
| -- | -- | -- | -- | -- |
| additionalRemarks |  | 1 | string |  |
| capturedWith |  | n | CAT/DeviceUsage |  |
| colorDepth |  | 1 | [emb] QuantitativeValue |  |
| compressionType | X | 1 | ControlledTerms terminology or ContentType or (Anaylsis)Technique?  | generally, lossy or loseless, but each can be achieved through different techniques, e.g., loseless - RLE vs. LZW |
| compressionRatio |  | 1 | [emb] QuantitativeValue | typically expressed as percentage of original image size with 100% = not resize |
| coordinateSpace | X | 1 | CustomCoordinateFramework or CommonCoordinateSpace | values don't exist yet and are still in discussion |
| dataLocation | X | 1 | File |  |
| dimension | X | 2 | integer |  |
| isPartOf |  | n | Image or Volume | 'shortcut' to describe the process of creating an image/volume from multiple images, e.g., tile scans or 3D reconstructions |
| lookupLabel |  | 1 | string |  |
| resolution | X | 2 | [emb] QuantitativeValue |  |

### VectorGraphic
| property | required? | count | value | comment |
| -- | -- | -- | -- | -- |
| additionalRemarks |  | 1 | string |  |
| coordinateSpace | X | 1 | CustomCoordinateFramework or CommonCoordinateSpace | values don't exist yet and are still in discussion |
| dataLocation | X | 1 | File |  |
| dimension | X | 2 | integer |  |
| lookupLabel |  | 1 | string |  |
| software |  | 1 | SoftwareVersion | existing property name used in computation extension, should be discussed |
| ??? |  |  |  | anything more? ChatGPT suggests detailed content descriptions like layer information (number of, names, order, visibility), color information (used model/palette or specific color codes), object properties (stroke width, opacity, etc.) and font/text information; this seems to extensive |

### ImageStack
NOTE: This is defined as 'A group of images with a similar frame of reference.' and include captured stacks (like z-stacks), reconstructed stacks from single image captures and images that could be reconstructed into a stack because they have a similar frame of reference (like 2D atlas reference data - data has same coordinate space and typically are single captures from the same brain at different spatial locations).
**QUESTION: When does a ImageStack become a Volume?**  
| property | required? | count | value | comment |
| -- | -- | -- | -- | -- |
| additionalRemarks |  | 1 | string |  |
| capturedWith |  | n | CAT/DeviceUsage |  |
| colorDepth |  | 1 | [emb] QuantitativeValue |  |
| compressionType | X | 1 | ControlledTerms terminology or ContentType or (Anaylsis)Technique?  | generally, lossy or loseless, but each can be achieved through different techniques, e.g., loseless - RLE vs. LZW |
| compressionRatio |  | 1 | [emb] QuantitativeValue | typically expressed as percentage of original image size with 100% = not resize |
| coordinateSpace | X | 1 | CustomCoordinateFramework or CommonCoordinateSpace | values don't exist yet and are still in discussion |
| dataLocation | X | n | File or Image or Volume? | not sure how to solve this |
| dimension | X | 2 or 3? | integer | if each image under 'dataLocation' has a dimension, should this be removed or reflect that information (in 2D) or reflect that information combined with the number of images (aka depth, in 3D)? |
| isPartOf |  | n | Volume | any point in having this here? |
| lookupLabel |  | 1 | string |  |
| numberOfImages | X | 2-n | integer |  |
| resolution | X | 2 or 3? | [emb] QuantitativeValue | same questions as for dimensions |
| sliceThickness or scanningDepth or StackDepth or depth |  | 1 | [emb] QuantitativeValue | not sure if 'sliceThickness' is the best name; e.g., a z-stack can be captured from a 30um thick tissue slice but total scanning depth might be 10um somewhere within that slice |
| imageSpacing or spacing |  | 1 | [emb] QuantitativeValue | with negative values indicating an overlap |

### Volume
| property | required? | count | value | comment |
| -- | -- | -- | -- | -- |
| additionalRemarks |  | 1 | string |  |
| capturedWith |  | n | CAT/DeviceUsage |  |
| colorDepth |  | 1 | [emb] QuantitativeValue |  |
| compressionType | X | 1 | ControlledTerms terminology or ContentType or (Anaylsis)Technique?  | generally, lossy or loseless, but each can be achieved through different techniques, e.g., loseless - RLE vs. LZW |
| compressionRatio |  | 1 | [emb] QuantitativeValue | typically expressed as percentage of original image size with 100% = not resize |
| coordinateSpace | X | 1 | CustomCoordinateFramework or CommonCoordinateSpace | values don't exist yet and are still in discussion |
| dataLocation | X | 1 | File |  |
| dimension | X | 3 | integer |  |
| isPartOf |  | n | Volume | similar 'shortcut' as in RasterGraphic, needed? |
| lookupLabel |  | 1 | string |  |
| resolution | X | 3 | [emb] QuantitativeValue |  |

### 3DMesh (one surface type)
NOTE: I know we need this somehow, but I'm not sure how to solve it since I don't really know this data type. 
| property | required? | count | value | comment |
| -- | -- | -- | -- | -- |
| additionalRemarks |  | 1 | string |  |
| compressionType | X | 1 | ControlledTerms terminology or ContentType or (Anaylsis)Technique?  | generally, lossy or loseless, but each can be achieved through different techniques, e.g., loseless - RLE vs. LZW |
| compressionRatio |  | 1 | [emb] QuantitativeValue | typically expressed as percentage of original image size with 100% = not resize |
| coordinateSpace | X | 1 | CustomCoordinateFramework or CommonCoordinateSpace | values don't exist yet and are still in discussion |
| dataLocation | X | 1 | File |  |
| dimension | X | 3 | integer |  |
| isPartOf |  | n | Volume | similar 'shortcut' as in RasterGraphic, needed? |
| lookupLabel |  | 1 | string |  |
| resolution | X | 1? | [emb] QuantitativeValue | as far as I can tell, this reflects the density of vertices which would be a single value? |
| software |  | 1 | SoftwareVersion | existing property name used in computation extension, should be discussed |
| vertexCount | X | 1 | integer |  |
| faceCount | X | 1 | integer |  |
| ??? |  |  |  | anything more? Seems like there are also texture coordinates and normals  which may or may not be important? Also, a faces is made up of a specific number of vertices, but as far as I understand this does not need to be uniform within a mesh (e.g., not always all triangles) but often is solved that way, should we have a property like '(mesh)FaceType' with a ControlledTerms terminology (e.g., triangular/triangle, quadrilateral/quad) and allow n or have a term 'mixed' in this terminology restricting it to exactly one; alternatively, it could be called 'faceDimension' with a list of or single integer (e.g., 3 would represent a triangle and [3, 4] woudl represent a mix of triangles and quads |

This is a rather long issue. @lzehl if this seems like a reasonable start point I can make this into a PR for easier review of each property. @openMetadataInitiative/openminds-admins and @openMetadataInitiative/openminds-developers, any feedback?

UlrikeS91:
adjusted draft of #523


UlrikeS91:
@Peyman-N, @apdavison and @lzehl I will keep it as a draft for now, but the content is ready for review. Except for one new ControlledTerms terminology*, everything should be functional in the existing framework. Meaning the coordinateSpace still points to the old setup. This will need to be updated when we do the SANDS updates. 

I restructured the content from the issue since it seems like several image types share properties (which makes sense). 

The schemas are structured the following way:

- **'Image'**: 
   - concept schema
   - contains info about coordinate space and file compression 
   - has a lookup label and additional remarks
   
- **'RasterBasedImage' and 'VectorBasedImage'**:
   - concept schemas
   - extend 'Image'
   - contain some similar properties: data location & dimension
   - but had to be split due to their inherent differences (vectors are scalable)
   - 'RasterBasedImage' additionally contains device used capture image, color depth and resolution
   - 'VectorBasedImage' additionally contains software used to create image 
   
- specific raster-based images restrict some of the inherited properties and define additional ones: 
   - 'RasterGraphic'
   - 'Volume'
   - 'ImageStack' 

- specific vector-based images restrict some of the inherited properties and define additional ones: 
   - 'VectorGraphic'
   - 'PolygonalMesh'

* new terminology is on 'Image', called 'ImageCompressionType' (e.g., LZW, RLE, JPEG), this could also be solved by defining the value as 'core:Technique' (or 'core:AnalysisTechnique', not sure) since this is basically what is happening - a method applied to the file to reduce the size of the file 

Looking forward to discussing this draft 🙂

lzehl:
to be further discussed:

2Draster
3Draster
2Dvector
3Dvector

compare to:
https://en.wikipedia.org/wiki/Scanning_probe_microscopy
https://fsl.fmrib.ox.ac.uk/fsl/oldwiki/FSLVBM.html

xgui3783:
Thanks @lzehl for pointing me to this PR. Apologies that I only have a chance to be involved now, and now sooner. I do have some questions to everyone involved:

1. Is there a way in the current proposal to include anchoring information? 

We currently work with the outputs of voluba (3 dimensional raster based image) and quicknii (2 dimensional raster based image). In both cases, through user interaction, a transform file is often produced (in the case of voluba, a 4x4 affine, in the case of quicknii, if I understand correctly, effectively a 3x3 matrix) 

In the current proposal `schemas/dataTypes/rasterBasedImage.schema.tpl.json` does not seem to allow the input of such a transform.

2. I want to clarify if it is possible - or if it is the intention of the proposal - to capture nonlinear transformations. 

In such a case, the relationship between the source (e.g. `rasterBasedImage`) to the target (`CommonCoordinateSpaceVersion|CustomCoordinateSpace`) is not defined by resolution, but a deformation field. 

I want to point out the effort done by the ngff v0.5[1] (which is an extension to zarr v3[2]), on how they handle the pixel/voxel --> physical unit problem[3]. They allow the definition of a list of transformation steps, which at the moment include scaling (not dissimilar to this proposal) and translation. In an array form, it lends itself to extension, should the initial proposal does not cover edge cases (e.g. rotation can be introduced via an additional transform type, nonlinear transform too)

Whilst ngff (and by extension, zarr) is mainly a data format primarly used to store raster 2d/3d image data, its approach to inform the client about how to transform from one space to another seems flexible. 

edit: 

ping @UlrikeS91 @Peyman-N @lzehl 


[1] https://ngff.openmicroscopy.org/latest/

[2] https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html

[3] https://ngff.openmicroscopy.org/latest/#multiscale-md

lzehl:
@xgui3783 thanks for getting involved here. Your input is needed and greatly appreciated. I will respond more in detail later but here already something that might clarify a bit our approach (so far)

These data type schemas suggested here would indeed not include anchoring information. It is not there purpose. Anchoring information would be stored in a separate schema (because there could be multiple to different coordinate spaces associated with the same file) which we should further discuss. The coordinate space in the here provided schemas will always refer to the coordinate space of the file (so the saved transformation result) not to the coordinate space a file could be transformed to (with the given transformation information). 

I would suggest to have a dedicated meeting for how we can extend SANDS with transformation activities and the respective results. We discussed this multiple times in the past but always dropped it for time and complexity reasons. 

@openMetadataInitiative/openminds-developers and @xgui3783 please provide additional comments/input to further shape this PR and related issues.

xgui3783:
> The coordinate space in the here provided schemas will always refer to the coordinate space of the file

Do I understand correctly then, the following examples **should** have the corresponding `commonCoordinateSpaceVersion` be set to their `.coordinateSpace` linked attribute :

- template image (e.g. T1/T2 MRI, 20um reconstruction big brain)
- parcellations (e.g. waxholm v1-4 in waxholm space, Julich Brain in colin/icbm152*, allen mouse 2015/2017 in allen ccf v3)
- other activation map/image/geometries (e.g. julich brain probability maps, output of fmriprep, nutil pointclouds etc)

But the following **should not** :

- volumes/images that needed more than "scaling" (e.g. Julich Brain delineations in Big Brain, 2D image anchored by QuickNii)

lzehl:
@xgui3783 to summarize: yes all images/parcellations that were registered to a common coordinate space version and saved in their transformed form should point to this common coordinate space version.

All images/parcellations that are saved in their original space, that are only accompanied by a file that provides the parameters/transformation matrix to perform an transformation to a common coordinate space version should not directly point to this common coordinate space version but to their native space. They need to be linked together with the transformation file through an activity to a common coordinate space version. This case we need to discuss. I think we have elements ready in our computation extension but should potentially have a specific extension for this in SANDS (maybe by creating a computation dependency @apdavison).

lzehl:
@UlrikeS91 sorry that we solved this now differently and many thanks for starting this. We solved this now in #579 for a first set of imaging data types and discussing in #585 the first set of time series data types.

PR #529, opened 2025-01-23
Update contentTypePattern.schema.tpl.json

lzehl:
Another typo fix that causes codespell to crash


lzehl:
@apdavison we would also update the version build to include this fix afterwards onto openMINDS v3.0

Raphael-Gazzotti (approved):

PR #530, opened 2025-01-30
Add an "entryPoint" property to ModelVersion as a way to improve specification of how to run simulations with a model

apdavison:
(cf the same property in openMINDS_computation/validationTestVersion)


lzehl (commented):
Is the entry point always expected to be a file path / IRI ? Then the value should be restricted accordingly (either by format or by pattern)

apdavison:
> Is the entry point always expected to be a file path / IRI ?

no, it could be an arbitrary string label.

lzehl (approved):

PR #535, opened 2025-02-06
Add a "configuration" property to ModelVersion 

apdavison:
as a way to improve specification of how to run simulations with a model.

(cf the same property in openMINDS_computation/validationTestVersion)

### DIFF
--- a/schemas/products/modelVersion.schema.tpl.json
+++ b/schemas/products/modelVersion.schema.tpl.json
@@ -6,6 +6,15 @@
     "license"
   ],
   "properties": {
+    "configuration": {
+      "_instruction": "Add the configuration information for this model version.",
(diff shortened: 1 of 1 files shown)


schemas/products/modelVersion.schema.tpl.json
```diff
@@ -6,6 +6,15 @@
     "license"
   ],
   "properties": {
+    "configuration": {
```
lzehl, line 9: should we make this rather a category?

apdavison: yes, that might be worthwhile, but I would like to merge this PR first, and then discuss adding a new category in a separate issue.

lzehl (approved):

PR #538, opened 2025-03-31
Fix JSON sent by curl in openMINDS_upstream.yml

apdavison:
The JSON document sent by curl to the central repository is missing a closing "}".

This may be why the commit messages in the central repo are missing information about the module and branch (e.g. "build triggered by submodule version" instead of "build triggered by submodule openMetadataInitiative/openMINDS_core version v5"


Raphael-Gazzotti:
Good catch! Will need to be also fixed for the other submodules (and versions).

Raphael-Gazzotti (approved):

apdavison:
Yes. I suggest doing this by direct push, rather than making a PR for each combination of submodule and branch. Do you agree @Raphael-Gazzotti @lzehl ?

apdavison:
ok, I've applied this fix to all the module x branch combinations, I think.

(I've also tweaked the build pipeline, which was still failing when triggered by a module change. I think everything is working as intended now).

Raphael-Gazzotti:
There is one thing that can be improved in the build pipeline, I will do a PR.

I added unnecessary code in "update_relevant_versions_from_repo", the "older" versions were already pulled by clone_central(True).

PR #541, opened 2025-04-04
Pipeline update v1

Raphael-Gazzotti:


lzehl (changes requested):
v1 of core is not functional with our pipeline. This was written in a pseudo language to capture that past status of the first KG version (which was not openminds).

lzehl:
@Raphael-Gazzotti I would just delete the workflows completely here.

Raphael-Gazzotti:
Not sync with the right branch, will create a new PR.

PR #553, opened 2025-06-17
switch connection between RP and RPVs

### LINKED ISSUE #515: Rediscuss direction of link between ResearchProduct and ResearchProductVersion
I would like to propose removing the property "hasVersion" from ResearchProduct and adding "isVersionOf" (or some similar name) to ResearchProductVersion**.

The reason is that with the current system, to add a new RPV requires a user to have write permissions for both the RPV and the RP, whereas if we reverse the direction, the user only needs to have write permissions for the RPV. The same goes when deleting a RPV.

In an EBRAINS context, this situation often arises when adding a new RPV (in a collab space) to an already-released RP (in a curator-only space), particularly for Models and LivePapers. The logic needed to manage all the possible situations, juggling a user KG client and a service client, gets very complicated. It would be greatly simplified if the link direction were reversed.

**(we might want to consider renaming "isNewVersionOf" and "isAlternativeVersionOf" at the same time, to avoid confusion, e.g. "isNewerThan" and "isAlternativeTo")

lzehl:
is implementation of #515 within the core sub-module. The issue also requires updates in SANDS (atlas + version, common coordinate space + version), computation (workflow recipe + version, validation test + version), and publication (live paper + version).

- deleted "hasVersion" from RPs
- added "isVersionOf" in RPVs linking to RP
- renamed "isNewVersionOf" in RPV to "precedingVersion"
- renamed "isAlternativeVersionOf" in PRV to "variantVersion"

### DIFF
--- a/schemas/products/dataset.schema.tpl.json
+++ b/schemas/products/dataset.schema.tpl.json
@@ -20,15 +20,6 @@
         "core:DOI",
         "core:IdentifiersDotOrgID"
       ]
-    },
-    "hasVersion": {
-      "type": "array",
-      "minItems": 1,
-      "uniqueItems": true,
-      "_instruction": "Add all versions of this dataset.",
-      "_linkedTypes": [
-        "core:DatasetVersion"
-      ]
     }
   }
 }
--- a/schemas/products/datasetVersion.schema.tpl.json
+++ b/schemas/products/datasetVersion.schema.tpl.json
@@ -91,7 +91,7 @@
       ]
     },
     "isVersionOf": {
-      "_instruction": "Add the dataset of this dataset version.",
+      "_instruction": "Add the version-independent information about this dataset.",
       "_linkedTypes": [
         "core:Dataset"
       ]
--- a/schemas/products/datasetVersion.schema.tpl.json
+++ b/schemas/products/datasetVersion.schema.tpl.json
@@ -75,6 +75,21 @@
         "https://openminds.ebrains.eu/sands/CommonCoordinateSpaceVersion"
       ]
     },
+    "isPrecededBy": {
+      "_instruction": "Add the dataset version preceding this dataset version.",
+      "_linkedTypes": [
+        "core:DatasetVersion"
+      ]
+    },
+    "isVariantOf": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add all dataset versions that can be used alternatively to this dataset version.",
+      "_linkedTypes": [
+        "core:DatasetVersion"
+      ]
+    },
     "isVersionOf": {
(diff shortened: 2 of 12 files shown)


lzehl:
@Raphael-Gazzotti I don't get why the schema-validator is still failing. Can you cross-check?

Raphael-Gazzotti:
@lzehl Apparently "inputs" is not supported with pull_request, I modified slightly the logic. Hopefully this works now.

lzehl:
@openMetadataInitiative/openminds-developers I realized that I have a mismatch in the renaming of the properties due to an open discussion that we did not finish:

RPV is now linked to RP with "isVersionOf". We need to decide:
- rename "isNewVersionOf" in RPV to "precedingVersion", "hasPrecedingVersion", or "isSuccessorOf" ??
- rename "isAlternativeVersionOf" in PRV to "variantVersion", "hasVariantVersion", or "isVariantOf" ??

UlrikeS91:
> * rename "isNewVersionOf" in RPV to "precedingVersion", "hasPrecedingVersion", or "isSuccessorOf" ??
> * rename "isAlternativeVersionOf" in PRV to "variantVersion", "hasVariantVersion", or "isVariantOf" ??

My opinion:
1. "isSuccessorOf" and "isVariantOf": For this purpose, I feel that having "version" in the property name would be better. It would make it more explicit. Although, for other purposes it may be useful to use these more ambiguous ones. I don't have a concrete example, though.
2. "precedingVersion" and "variantVersion": Nice and short, but could be misleading. 
3. "hasPrecedingVersion" and "hasVariantVersion": The action word removes the ambiguity from point 2. 

If we would like to have the option of using the properties for other things than "versions", I'm leaning towards point 1. Otherwise, I would choose point 3. 

@apdavison What would be your take on this?

apdavison:
With an eye to developing a formal ontology in the future, I think less-explicit properties are better, as they capture commonalities between things of different types.

So I like `isSuccessorOf` (or `isPrecededBy`, which I think I prefer because it just refers to temporal ordering, and has less of an implication of "replaces") and `isVariantOf` or `hasVariant` (since this relationship is bi-directional).

We could go a step further, and use `follows` for the previous version (or previous whatever).

lzehl:
Discussion output from openMINDS dev meeting:
- isPrecededBy
- isVariantOf

lzehl:
@apdavison updated now to latest decisions (I hope) and ready for your review :)

apdavison (approved):

apdavison:
When I merged this, I forgot to check the target. This should have been merged into v6, not v5 (which is the version used in openMINDS v4.0).

I have merged these changes into v6, and reverted v5 back to its condition before the merge.

Issue #330, opened 2022-04-21
request for open value ranges


lzehl:
@HumanBrainProject/openminds-developers we frequently get the request to allow for open ranges specifically for age information.
E.g. indication that all participants were at least 18 years old or older.

I see three options how we can solve this within the schemas:

1) make both minimum value and maximum value in the QuantitativeValueRange schema optional (less preferred option from my side)
2) introduce a new schema QuantitativeValueRangeOpen that has the properties "border value" (e.g., 18), "relational operator" (e.g., "≥"; should be controlledTerm), and "unit" (e.g., "year"; is controlledTerm) (seems to me the fastest clean solution for this issue)
3) extend the openMINDS syntax for schemas to define ONE-OF required properties and change the QuantitativeValueRange schema accordingly (this would give us the most flexibility also for other cases with the same issue, but technically it's not so easy to solve as far as I can see, because the one-of option in json-schema seems not to be build for required properties)

What do you think? (comments from outside of the openMINDS development team are of course also welcome)

UlrikeS91:
3.) would be best, but I understand the issue. If we need it asap, then I would go for 2.) and not for 1.)

jcolomb:
what about having a value set for this purpose in the QuantitativeValueRange schema, so you can set minimum value 18, maximum value of "undefined" or "right-censored-data" or some other term you would find appropriate?

It might make sense for other kind of range where the maximum is only the maximum we are looking at (example max time to find a platform in a water maze experiment)

Peyman-N:
I really prefer option 3 too.
But I fail to see why if we do it with `ONE-OFF` would cause any problem. For example this should work sufficiently enough: 
```
"maxValue": {
    "anyOf":[
        {
            "type": "number"
        },
        {
            "const" : "Infinity",
        },
        "_instruction": "Add the maximum value measured for this range, if this is an open range specify 'Infinity'."
    }
```
Dose it cause a problem with generator ?

lzehl:
@Peyman-N & @jcolomb thanks for the suggestion but the problem here is that you have to defined it for both ranges which could then lead to the following: (infinity to infinity), (undefined to undefined), (undefined to infinity). 

With the "one-off" option, I meant a schema definition where you have to state either a minValue, or a maxValue or both which I think is not possible at the moment.

lzehl:
And: string OR number will cause issues with editors (at least at the moment with the KG Editor)

Peyman-N:
[...]

PR #556, opened 2025-07-16
Add QuantitativeValueRangeUnbounded schema template.

Raphael-Gazzotti:


apdavison:
see my comment on #330

lzehl:
see ongoing / updated discussion on #330 (goal: option 3 with implementation after v5.0)

PR #563, opened 2025-09-29
Remove WebService and WebServiceVersion, replace with Service, Interface and InterfaceDependency

apdavison:
[Design document v0.3](https://notes.ebrains.eu/s/5mLA5Bi7H#)
[Design document v0.4](https://notes.ebrains.eu/0EsVHXF4ToKip1rt-JP6Ow)

Related:
- https://github.com/openMetadataInitiative/openMINDS_core/issues/528
- https://github.com/openMetadataInitiative/openMINDS_core/issues/527
- https://github.com/openMetadataInitiative/openMINDS_computation/pull/62

Replaces:

https://github.com/openMetadataInitiative/openMINDS_computation/pull/63


schemas/products/interface.schema.tpl.json
```diff
@@ -0,0 +1,73 @@
...
+      "_instruction": "Enter a web resource that defines this interface, e.g., an OpenAPI.json document.",
+      "_linkedTypes": [
+        "core.WebResource"
+      ]
+    },
+    "description": {
+      "type": "string",
+      "_instruction": "Enter a description of this interface."
+    },
+    "developer": {
```
lzehl: This should be contributor according to our new plans. However we need to decide for one of the suggested options first https://github.com/openMetadataInitiative/openMINDS_core/issues/513

schemas/products/interface.schema.tpl.json
```diff
@@ -0,0 +1,73 @@
...
+    "versionIdentifier"
+  ],
+  "properties": {
+    "communicationProtocol": {
+      "_instruction": "Enter the communication protocol used by this interface.",
+      "_linkedTypes": [
+        "controlledTerms:CommunicationProtocol"
+      ]
+    },
+    "custodian": {
```
lzehl: Should most like be removed depending on the option decided in https://github.com/openMetadataInitiative/openMINDS_core/issues/513

schemas/products/interface.schema.tpl.json
lzehl, line 1: If interfaces are versioned they should in principle be handled as research products with research products versions. However I though we wanted to conceptualize the interface, and assume the code base is the software (making the interface explicitly not a research product with independent versioning). Otherwise the link to the code is missing that would allow for an independent versioning. 

Question: If the same software version code base implements multiple versions of the same interface should we just treat them as equal variants with no preference? Then the version identifier actually just becomes just an internal identifier (no version specification needed, should be covered in definition and description). If there is a preference then we need to treat the interface as an independent product with versions I think (we may want to rethink the base elements of RP/RPV). 

Do we need to adapt the naming here as well to CommunicationInterface ?

schemas/products/interface.schema.tpl.json
lzehl: Also: should the developer not be part of the SoftwareVersion rather then on the Interface? even if code is only adjusted for the interface the software version will have to change (I though we wanted to conceptualize the Interface, and assume the code base is the software).

schemas/products/service.schema.tpl.json
```diff
@@ -0,0 +1,118 @@
...
+      "_linkedTypes": [
+        "core:Protocol",
+        "computation:WorkflowRecipe"
+      ]
+    },
+    "description": {
+      "type": "string",
+      "_instruction": "Enter a description (or abstract) of this service."
+    },
+    "feature": {
```
lzehl: Since Services are now broader this can't be only pointing to SoftwareFeatures

schemas/products/service.schema.tpl.json
```diff
@@ -0,0 +1,118 @@
+{
+  "_type": "core:Service",
+  "required": [
+    "fullDocumentation",
+    "fullName",
+    "shortName",
+    "type"
+  ],
+  "properties": {
+    "custodian": {
```
lzehl: should be removed depending on what we decide for https://github.com/openMetadataInitiative/openMINDS_core/issues/513

schemas/products/service.schema.tpl.json
lzehl, line 1: Needs to explicitly links to the legal entity providing the service (through https://github.com/openMetadataInitiative/openMINDS_core/issues/513 or specifically). We need to be able to filter services for this relation (custodian might in principle be also from somewhere else)

schemas/products/interfaceDependency.schema.tpl.json
lzehl, line 1: If in the software version the dependency can be to an interface or to a software version this can't be interface specific to interface but needs to be for interface or software version (failure impact would matter for both, no?)

schemas/products/softwareVersion.schema.tpl.json
lzehl, line 1: SoftwareApplicationCategory list and SoftwareFeature list need to be revised I think

schemas/products/softwareVersion.schema.tpl.json
```diff
@@ -138,12 +160,12 @@
...
         "controlledTerms:ProgrammingLanguage"
       ]
-    },    
+    },
     "requirement": {
       "type": "array",
       "minItems": 1,
       "uniqueItems": true,
-      "_instruction": "Enter all requirements of this software version.",
+      "_instruction": "Enter any requirements of this software version that are not listed under the 'dependency' property.",
```
lzehl: Please explain difference between requirement and dependency. I don't see why we want to separate this. Could we think of a joint solution?

lzehl (changes requested):

schemas/products/interface.schema.tpl.json
apdavison: An interface is independent of the software that implements it. It is possible to have different software tools that implement the same interface.

schemas/products/interface.schema.tpl.json
apdavison: An interface is an independent product. There can be multiple pieces of software that implement a given interface; equally a given piece of software can implement multiple versions of a given interface. There is not necessarily any relationship between the interface version identifier and the software version identifier.

I'm not sure it should be considered a _research_ product, however.

schemas/products/service.schema.tpl.json
apdavison: agreed, this needs some discussion.

schemas/products/softwareVersion.schema.tpl.json
apdavison: agreed

schemas/products/interfaceDependency.schema.tpl.json
apdavison: I'm not sure I understand, can you clarify?

schemas/products/softwareVersion.schema.tpl.json
apdavison: I did it this way for backwards compatibility and ease of migration. "requirement" is an array of strings, while "dependency" is an array of instances. My idea was to keep them both for v5, and then merge them for v6. Otherwise, we need to create SoftwareVersions for all strings in existing "requirement" properties during the v4-v5 migration.

schemas/products/interface.schema.tpl.json
lzehl: So where is the code for the interface? we don't link it right now

schemas/products/interface.schema.tpl.json
lzehl: We should not be concerned about the semantics here. The research product can easily be renamed to something more general (would anyway make sense). 

If it is an independent product it needs to be implemented in the same way as the others and it has to have then it's own code base.

However I understood this a bit differently. If we have code this is first and foremost a software. That software might have code for an interface as well or only holds the code for an interface, but the interface itself is just the concept of what the code is encoding. 

@olinux can we please get your input here as well.
@apdavison maybe some examples would help for the expected cases?

apdavison: For REST APIs, the interface definition is more like a metadata model. It is very common to use the OpenAPI standard (formerly Swagger), for which the interface definition is a single JSON file. I don't think we would register the openMINDS schema templates as software?

For non-REST APIs, the interface definition would probably be a human-readable, but not machine-readable document.

apdavison: (Here is the definition for the KG Core API: https://core.kg.ebrains.eu/v3/api-docs/v3)

apdavison: As another example, at one point I rewrote the neuromorphic Job Queue API from scratch (for maintainability and performance), using different Python libraries but keeping the same API.

schemas/products/interface.schema.tpl.json
apdavison: The "definition" property links to a WebResource, which is generally a JSON file for REST APIs, or could be a formal human-readable document for other kinds of interface.

schemas/products/softwareVersion.schema.tpl.json
lzehl: I don't think we need to create SoftwareVersions. We can also say a property called "requirement" or "dependency" to embed ToolDependency (name TBD) which states the failureImpact and links to either an Interface, a SoftwareVersion, a File (to requirements.txt; TBD), a WebResource (to requirements.txt; TBD) or a Configuration (or a StringProperty).

Let's discuss this further

schemas/products/interface.schema.tpl.json
apdavison: "author" would be a better term than "developer" (or rather, we should use the "author" role when we switch to "contributor")

schemas/products/interfaceDependency.schema.tpl.json
lzehl: aims towards the issue I have with having a "dependency" and a "requirement" property in the software version. impact of failure is critical in both cases.

schemas/products/interfaceVersion.schema.tpl.json
lzehl, line 1: would an interfaceVersion ever specify a repository?

schemas/products/service.schema.tpl.json
```diff
@@ -0,0 +1,94 @@
...
+      "_linkedCategories": [
+        "publicationReference"
+      ]
+    },
+    "scope": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add the scope of this service, i.e., what analysis methods does it perform, what kinds of experimental data is it intended for.",
+      "_linkedTypes": [
```
lzehl: should become linked category toolScope

schemas/products/service.schema.tpl.json
```diff
@@ -0,0 +1,94 @@
...
+        "keyword"
+      ]
+    },
+    "relatedPublication": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "_instruction": "Add all further publications besides the full documentation that provide the original context for the production of this service (e.g., an original research article describing the service).",
+      "_linkedCategories": [
+        "publicationReference"
```
lzehl, line 55: is this introduced?

schemas/products/service.schema.tpl.json
```diff
@@ -0,0 +1,94 @@
...
+      "_instruction": "Enter all channels through which a user can receive support for using this service.",
+      "items": {
+        "type": "string",
+        "_formats": [
+          "email",
+          "iri"
+        ]
+      }
+    },
+    "type": {
```
lzehl: if I remember correctly we said that we want to remove this for now?

schemas/products/softwareVersion.schema.tpl.json
```diff
@@ -4,41 +4,14 @@
   "_categories": [
     "setupComponent"
   ],
-  "required": [
```
lzehl, line 7: seems a bit weird to not require anything we want to have for software version. isn't there one thing we really want to keep (like scope)?

schemas/products/softwareVersion.schema.tpl.json
```diff
@@ -49,25 +22,13 @@
...
       "type": "array",
       "minItems": 1,
       "uniqueItems": true,
-      "_instruction": "Add all distinguishing characteristics of this software version (e.g., performance, portability or functionality).",
+      "_instruction": "Add all interface versions that are implemented by this software version.",
       "_linkedTypes": [
-        "controlledTerms:SoftwareFeature"
-      ]
-    },
-    "hasPart": {
```
lzehl, line 60: I think it is correct to remove this from the SoftwareVersion. 

hasPart SoftwareVersion is now handled through dependency.

all other relations rather belong to the deployment of the software version since these are configuration inputs (that can be used or not). so this belongs to the new deployment schema

schemas/products/softwareVersion.schema.tpl.json
```diff
@@ -144,15 +105,19 @@
...
+    "scope": {
       "type": "array",
       "minItems": 1,
       "uniqueItems": true,
-      "_instruction": "Enter all requirements of this software version.",
-      "items": {
-        "type": "string"
-      }
+      "_instruction": "Add terms that describe what this software version does.",
+      "_linkedTypes": [
```
lzehl: should link to the category toolScope

lzehl (approved):
needs to be merged after setting up the toolScope category (with the missing controlledTerm)

PR #573, opened 2025-12-11
Closes #572: ISBN pattern extension

### LINKED ISSUE #572: ISBN pattern needs to be updated
I just noticed that our regex pattern for ISBNs is not inclusive enough. For example, the Paxinos Rat Atlas 6th edition's ISBN is 0-12-547612-4. 

If I'm not mistaken, our pattern (^([0-9]{3}-|)[0-9]{1}-[0-9]{3}-[0-9]{5}-[0-9]{1}$) enforces a (3 digits -) 1 digit - 3 digits - 5 digits - 1 digit pattern (i.e., X-XXX-XXXXX-X (10-digit ISBN) or XXX-X-XXX-XXXXX-X (13-digit ISBN)), but the ISBN above follows a 1 digit - 2 digits - 6 digits - 1 digit pattern.  

I'm not 100% comfortable with proposing any changes to the regex (because it's a regex...). Could somebody fix this? @Raphael-Gazzotti maybe? 

I'm not sure if this also affects ISSNs, but we should probably investigate. 

FYI @lzehl

Raphael-Gazzotti:
Extends original pattern matching to ISBN that follows a 1 digit - 2 digits - 6 digits - 1 digit pattern.

### DIFF
--- a/schemas/digitalIdentifier/ISBN.schema.tpl.json
+++ b/schemas/digitalIdentifier/ISBN.schema.tpl.json
@@ -9,7 +9,7 @@
   "properties": {
     "identifier": {
       "type": "string",
-      "pattern": "^([0-9]{3}-|)[0-9]{1}-[0-9]{3}-[0-9]{5}-[0-9]{1}$",
+      "pattern": "^(([0-9]{3}-)?[0-9]{1}-[0-9]{3}-[0-9]{5}-[0-9]{1}|[0-9]{1}-[0-9]{2}-[0-9]{6}-[0-9]{1})$",
       "_instruction": "Enter the numeric commercial book identifier 'International Standard Book Number' (ISBN) following the defined pattern (e.g., 123-4-567-89012-3 (13-digit ISBN) or 4-567-89012-3 (10-digit ISBN))."
     }
   }
--- a/schemas/digitalIdentifier/ISBN.schema.tpl.json
+++ b/schemas/digitalIdentifier/ISBN.schema.tpl.json
@@ -9,7 +9,7 @@
   "properties": {
     "identifier": {
       "type": "string",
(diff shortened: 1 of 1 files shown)


schemas/digitalIdentifier/ISBN.schema.tpl.json
lzehl, line 1: I think the regex pattern is not yet fully correct.

Seems the setting for 13 digit codes can be (sum of digits must be always 13):
prefixElement(3 digits)-registrationGroupElement(1–5 digits)-registrantElement(1-7 digits)-publicationElement(1-6 digits)-checkDigit(1 digit)

Seems the setting for 10 digit codes can be (sum of digits must be always 10):
registrationGroupElement(1–5 digits)-publicationElement(2–7 digits)-titleIdentifier(1–6 digits)-checkDigit(1 digit)

Not really sure how to solve this.

lzehl (changes requested):

schemas/digitalIdentifier/ISBN.schema.tpl.json
Raphael-Gazzotti: Apparently there are some additional rules detailed here https://en.wikipedia.org/wiki/ISBN:
For the 10 digit codes: the final 'digit' can be a X.
For the 13 digit codes: they are always starting by 978- or 979-

Raphael-Gazzotti: Unfortunately we can't check the checksum with only a REGEX.

Raphael-Gazzotti: The pattern we discussed appears to be correct.

For ISBN-10, the only formal specification I was able to locate is:
https://www.barcodes4u.com.au/download/G1.pdf

For ISBN-13, specifications may vary depending on the source. However, the official ISBN International Users’ Manual (7th edition) states that the Registrant element may be up to 7 digits:
https://web.archive.org/web/20191211194637/https://www.kb.se/download/18.71dda82e160c04f1cc412bc/1531827912246/ISBN%20International%20Users%20Manual%20-%207th%20edition.pdf

Some secondary sources describe the ISBN-13 Registrant element as ranging from 2 to 7 digits, but this limitation does not appear in the official documentation.

Raphael-Gazzotti:
For testing purpose:
https://regex101.com/r/zeyPdS/1

lzehl (approved):

PR #582, opened 2026-02-23
Adding localFile schema from computation

Alixbonard:
Related issue: https://github.com/openMetadataInitiative/openMINDS_core/issues/557

Move localFile schema from computation to core
@lzehl could you check it ?


lzehl (approved):

lzehl (changes requested):
@Alixbonard should go into v6 branch not v5 branch

PR #590, opened 2026-03-03
Move recording and channels from ephys to core

Alixbonard:
related issue: https://github.com/openMetadataInitiative/openMINDS_core/issues/585 

- [x] removed recordedWith (see issue https://github.com/openMetadataInitiative/openMINDS_core/issues/588)
- [x] harmonized instructions and linkedType for regularTimeSeries 

@lzehl Should I do something more here ?

### DIFF
--- a/schemas/data/channel.tpl.json
+++ b/schemas/data/channel.tpl.json
@@ -7,7 +7,7 @@
   "properties": {
     "internalIdentifier": {
       "type": "string",
-      "_instruction": "Enter the identifier (or label) of this channel that is used within the corresponding data files to identify this channel."
+      "_instruction": "Enter the identifier (or label) used within the corresponding data files to uniquely identify this channel."
     },
     "unit": {
       "_instruction": "Add the unit of measurement for this channel.",
--- a/schemas/data/channel.tpl.json
+++ b/schemas/data/channel.tpl.json
@@ -1,0 +1,19 @@
+{
+  "_type": "core:Channel",
+  "required": [
+    "internalIdentifier",
+    "unit"
+  ],
+  "properties": {
+    "internalIdentifier": {
+      "type": "string",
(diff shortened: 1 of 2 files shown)


schemas/data/channel.tpl.json
```diff
@@ -0,0 +1,19 @@
+{
+  "_type": "core:Channel",
+  "required": [
+    "internalIdentifier",
+    "unit"
+  ],
+  "properties": {
+    "internalIdentifier": {
+      "type": "string",
+      "_instruction": "Enter the identifier (or label) of this channel that is used within the corresponding data files to identify this channel."
```
lzehl:
```suggestion
      "_instruction": "Enter the identifier (or label) used within the corresponding data files to uniquely identify this channel."
```

schemas/data/regularTimeSeries.tpl.json
```diff
@@ -0,0 +1,50 @@
...
+    },
+    "internalIdentifier": {
+      "type": "string",
+      "_instruction": "Enter the identifier (or label) of this regular time series that is used within the corresponding data files to identify this regular time series."
+    },
+    "name": {
+      "type": "string",
+      "_instruction": "Enter a descriptive name for this regular time series."
+    },
+    "previousRegularTimeSeries": {
```
lzehl, line 37: For what is this used? I would really like to remove it unless it is already in use.

lzehl (changes requested):

schemas/data/regularTimeSeries.tpl.json
Alixbonard: So far, in the in-depth curation workflow, previousRegulrTimeSeries (which corresponds to previousRecording) is not used.

lzehl (approved):

